PR #7, opened 2018-09-05
Handle network CIDR properly

vikaschoudhary16:


enxebre:
This is an improvement over what we have.
Caveats:
- This is not releasing ips, if the machine controller gets restarted for any reason there'll be overlap
- Ideally we should be able to instantiate a "libvirt manager" object when creating a new actuator and bury the offset variable and the logic around it at the libvirt manager level
- We should be able to put somenthing dynamic in place similar to 
https://github.com/dmacvicar/terraform-provider-libvirt/blob/2150005da77ddc623992685a8a574c8c170c37e6/libvirt/resource_libvirt_domain.go#L1401-L1423

PR #10, opened 2018-09-10
add libvirt dependency to Makefile

paulfantom:
Solution to run everything in containers (needed for CI).

Examples are implemented in https://github.com/openshift/release/pull/1415

/CC: @enxebre @petr-muller

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ depend-update:
 	dep ensure -update
 
 libvirt:
-	@if [ $(CONTAINERIZED) != 1 ]; then yum install -y libvirt-devel; fi
+	@if [ $(CONTAINERIZED) == 1 ]; then yum install -y libvirt-devel; fi
 
 generate: gendeepcopy
 
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ depend:
 depend-update:
 	dep ensure -update
 
-libvirt:
+deps-cgo:
 	@if [ $(CONTAINERIZED) == 1 ]; then yum install -y libvirt-devel; fi
 
 generate: gendeepcopy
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -43,14 +48,15 @@ push:
 
 check: fmt vet
 
-test:
+test: libvirt
```
enxebre: You are setting CONTAINERIZED=1 here https://github.com/openshift/release/pull/1415/files#diff-8494a5bc314f945b17df41d8bb54cfc9R32
So CGO_deps won't execute so woulnd't this still fail?

Makefile
```diff
@@ -16,13 +16,18 @@
...
+CONTAINERIZED ?= 0
+
 depend:
 	dep version || go get -u github.com/golang/dep/cmd/dep
 	dep ensure
 
 depend-update:
 	dep ensure -update
 
+libvirt:
```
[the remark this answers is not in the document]
paulfantom: why so convoluted naming scheme?

Makefile
paulfantom: You are right. I fixed condition so it should work now.

Makefile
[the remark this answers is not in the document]
paulfantom: I changed it to sth more in line with the rest of Makefile

Makefile
```diff
@@ -16,13 +16,18 @@
...
+CONTAINERIZED ?= 0
+
 depend:
 	dep version || go get -u github.com/golang/dep/cmd/dep
 	dep ensure
 
 depend-update:
 	dep ensure -update
 
+deps-cgo:
```
enxebre: why CONTAINERIZED?

PR #11, opened 2018-09-10
use official rh tooling image

paulfantom:
Same as in https://github.com/openshift/cluster-api-provider-aws/pull/40


paulfantom:
@derekwaynecarr unit tests are failing here due to config. Solution is created in https://github.com/openshift/cluster-api-provider-libvirt/pull/10

PR #22, opened 2018-09-21
add terraform to provision packet machine

paulfantom:
Add a script to provision packet machine and configure it so we can connect to it with libvirt client via SSH or TCP. It also automatically creates temporary ssh key file if needed (default behavior). 

/wip

### DIFF
--- a/hack/packet-provision.sh
+++ b/hack/packet-provision.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+set +e
+
+# Your Packet user account
+if [ "$PACKET_AUTH_TOKEN" == "" ]; then
+    echo "You need to set PACKET_AUTH_TOKEN variable first."
+    echo "Make sure that your SSH key is also set in packet.net"
+    exit 1
+fi
+
+export TF_VAR_environment_id=${ENVIRONMENT_ID:-$(uuidgen | cut -c1-8)}
+
+cd ./prebuild
+case ${1} in
+  "install")
+    ssh_path="$TF_VAR_ssh_key_path"
+    if [ "$TF_VAR_ssh_key_path" == "" ]; then
+        echo -e "\e[33mCreating temporary SSH file\e[0m"
+        ssh-keygen -t rsa -b 4096 -C "temporary packet.net key" -P "" -f "/tmp/packet_id_rsa" -q
+        ssh_path="/tmp/packet_id_rsa.pub"
+    fi
+    terraform init -input=false
+    terraform plan -input=false -out=tfplan.out && terraform apply -input=false -auto-approve tfplan.out
+    echo -e "\e[32m"
+    echo -e "*** Your packet.net host is called ${TF_VAR_environment_id}"
+    echo -e "*** You can also access it via SSH with key located in ${ssh_path}"
+    echo -e "\e[0m"
+    ;;
+  "destroy")
+    terraform destroy -input=false -auto-approve
+    rm /tmp/packet_id_rsa* || :
+    ;;
+  *)
+    echo "Use '$0 install' or '$0 destroy'."
+    ;;
+esac
--- a/hack/packet-provision.sh
+++ b/hack/packet-provision.sh
@@ -8,7 +8,13 @@ if [ "$PACKET_AUTH_TOKEN" == "" ]; then
     exit 1
 fi
 
-export TF_VAR_environment_id=${ENVIRONMENT_ID:-$(uuidgen | cut -c1-8)}
+# Your Packet user account
+if [ "$TF_VAR_packet_project_id" == "" ]; then
+    echo "You need to set TF_VAR_packet_project_id variable first."
+    exit 1
+fi
+
+export TF_VAR_id=${ID:-$(uuidgen | cut -c1-8)}
 
 cd ./prebuild
 case ${1} in
@@ -17,7 +23,7 @@ case ${1} in
     if [ "$TF_VAR_ssh_key_path" == "" ]; then
         echo -e "\e[33mCreating temporary SSH file\e[0m"
         ssh-keygen -t rsa -b 4096 -C "temporary packet.net key" -P "" -f "/tmp/packet_id_rsa" -q
-        ssh_path="/tmp/packet_id_rsa.pub"
+        ssh_path="/tmp/packet_id_rsa"
     fi
     terraform init -input=false
     terraform plan -input=false -out=tfplan.out && terraform apply -input=false -auto-approve tfplan.out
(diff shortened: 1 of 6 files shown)


hack/prebuild/main.tf
```diff
@@ -0,0 +1,32 @@
+resource "packet_project" "libvirt_actuator" {
+  name = "libvirt-actuator tests"
+}
+
+resource "packet_ssh_key" "key" {
+  name       = "unlikely_tf_ssh_key_name"
```
spangenberg: You should put ${var.environment_id} here.

hack/prebuild/main.tf
```diff
@@ -0,0 +1,32 @@
+resource "packet_project" "libvirt_actuator" {
+  name = "libvirt-actuator tests"
```
spangenberg: Maybe put ${var.environment_id} here.

hack/prebuild/main.tf
```diff
@@ -0,0 +1,32 @@
...
+}
+
+resource "packet_ssh_key" "key" {
+  name       = "unlikely_tf_ssh_key_name"
+  public_key = "${file("${var.ssh_key_path}")}"
+}
+
+resource "packet_device" "libvirt" {
+  hostname         = "${var.environment_id}"
```
spangenberg: Prepend something non alpha numeric, in case uuid generates a number fist.

hack/prebuild/init.sh
```diff
@@ -0,0 +1,19 @@
...
+unix_sock_rw_perms = "0770"
+listen_tls = 0
+listen_tcp = 1
+auth_tcp="none"
+tcp_port = "16509"
+EOF
+
+echo 'LIBVIRTD_ARGS="--listen"' >> /etc/sysconfig/libvirtd
+
+iptables -I INPUT -p tcp --dport 16509 -j ACCEPT -m comment --comment "Allow insecure libvirt clients"
```
bison: Are the tests going to use TCP or SSH as a transport for libvirt? I don't think it's okay to run these with unauthenticated libvirt listening on the public internet.

hack/prebuild/main.tf
```diff
@@ -0,0 +1,32 @@
+resource "packet_project" "libvirt_actuator" {
```
bison: Does this create a project for every build? We probably want a single re-used project. That will let us manage permissions to for it.

hack/prebuild/init.sh
enxebre: For CI I was thinking on parameterising the ip to allowed to access and then reuse your tls stuff, so we test it as it's consumed by the installer/mao. wdyt?

bison: Yeah, that sounds good. I'm not familiar with the e2e tests or the plan for those. Do they use the installer? If so, I think this should be on-hold until https://github.com/openshift/installer/pull/296 merges.

enxebre: Eventually we might use installer. Currently for mao/aws-actuator we use terraform to create the minimal infra needed to satisfy some actuator assumptions and verify expected behaviour, 
See https://github.com/openshift/machine-api-operator/tree/master/tests
e.g The minimal infra for libvirt actuator will contain a network and the ign volumes

bison: Ah, cool. Then I think it should be pretty easy to use TLS or SSH as the transport from the start. There aren't really any changes required for the actuator. The credentials, certs or SSH keys, just need to be available and the right URI needs to be used. The libvirt library will handle the rest.

bison: Just to add a little more detail on making TLS work: [here's a link to the docs](https://github.com/openshift/installer/blob/41a24535cb6025827ff4976650424d439e8c6661/docs/dev/libvirt-howto.md#17-configure-libvirt-to-accept-tls-connections)

We just need to generate the TLS assets somehow -- could use certtool, cfssl, or even Terraform if we don't want to use the Go tool in the installer PR. Then configure libvirtd to point to the certs and use a URI like: `qemu+tls://$IP/system?pkipath=/path/to/certs`

That should be enough and pretty secure as long as no one can access the generated certs and they are one-time use.

paulfantom: It would be probably easier to adapt it to use SSH as this already need SSH keys to connect to instance.

bison: Totally fine. As long as we're doing authentication and encryption somehow.

hack/packet-provision.sh
```diff
@@ -0,0 +1,36 @@
...
+case ${1} in
+  "install")
+    ssh_path="$TF_VAR_ssh_key_path"
+    if [ "$TF_VAR_ssh_key_path" == "" ]; then
+        echo -e "\e[33mCreating temporary SSH file\e[0m"
+        ssh-keygen -t rsa -b 4096 -C "temporary packet.net key" -P "" -f "/tmp/packet_id_rsa" -q
+        ssh_path="/tmp/packet_id_rsa.pub"
+    fi
+    terraform init -input=false
+    terraform plan -input=false -out=tfplan.out && terraform apply -input=false -auto-approve tfplan.out
```
frobware: As we have `-e` this could be two separate commands.

hack/packet-provision.sh
```diff
@@ -0,0 +1,36 @@
...
+    echo -e "\e[32m"
+    echo -e "*** Your packet.net host is called ${TF_VAR_environment_id}"
+    echo -e "*** You can also access it via SSH with key located in ${ssh_path}"
+    echo -e "\e[0m"
+    ;;
+  "destroy")
+    terraform destroy -input=false -auto-approve
+    rm /tmp/packet_id_rsa* || :
```
frobware: Why not use `rm -f`, instead of `||`?

hack/prebuild/init.sh
```diff
@@ -0,0 +1,19 @@
+#/bin/bash
+
+yum install -y -d1 libvirt libvirt-daemon-kvm
+usermod -aG libvirt root
```
frobware: Not sure I've looked to see how this script is used (or when) but do you not need `newgrp libvirt` here having modified the groups?

hack/prebuild/init.sh
```diff
@@ -0,0 +1,19 @@
...
+listen_tcp = 1
+auth_tcp="none"
+tcp_port = "16509"
+EOF
+
+echo 'LIBVIRTD_ARGS="--listen"' >> /etc/sysconfig/libvirtd
+
+iptables -I INPUT -p tcp --dport 16509 -j ACCEPT -m comment --comment "Allow insecure libvirt clients"
+
+systemctl start libvirtd
```
frobware: Do we also need `systemctl enable libvirtd`? Is there any expectation of this starting on reboot?

paulfantom: No, those machines aren't expected to be rebooted.

hack/prebuild/init.sh
paulfantom: It is used to install libvirt on packet host the easiest way.

hack/packet-provision.sh
paulfantom: Right, I will fix that in a follow-up PR

PR #27, opened 2018-09-28
terraform code for local libvirt testing

trawler:


paulfantom:
Ok so all this is basically one bash and one virsh command, right? This:
```bash
cat <<EOF > new_net.xml
<network>
  <name>tt.testing</name>
  <bridge name="tt0" />
  <forward mode="route" />
  <ip address="192.168.124.1" netmask="255.255.255.0" />
</network>
EOF
net-create --file new_net.xml
```
if it is, why do we need whole terraform for this, especially when we also need to execute some commands manually?

trawler:
@paulfantom

trawler:
@paulfantom the only manual commands that are essential is the dnsmasq one.
As for using bash scripts to set-up libvirt, etc, that can be said in many uses of terraform. I believe that in the end, terraform provides a simpler, cleaner way to quickly erect and destroy ad-hoc set-ups, and we can also automate the dnsmasq entries. @frobware  @enxebre

paulfantom:
I don't have anything against terraform. But do you really need to create and destroy this setting so many times? Is this something that you set up and leave? Do we really need the overhead and complexity of terraform just to locally create a network bridge and a couple of volumes?

As for automating dnsmasq entry, we shouldn't even touch NetworkManager config with automation. It might cause problems when people have more complex setups.

hack/prebuild/local-dev/README.md
```diff
@@ -0,0 +1,37 @@
+# Setting up a local libvirt dev environment
+
+## Set up NetworkManager DNS overlay
+_This step is optional, but useful for being able to resolve cluster-internal hostnames from your host._
+
```
enxebre: Don't think we need dns resolution for now, we we might wanna skip all this?

hack/prebuild/local-dev/README.md
```diff
@@ -0,0 +1,37 @@
...
+For this example:
+```
+> echo server=/tt.testing/192.168.124.1 | sudo tee /etc/NetworkManager/dnsmasq.d/tectonic.conf
+> systemctl restart NetworkManager
+```
+## Install the terraform provider
+Make sure you have the virsh binary installed: `sudo dnf install libvirt-client libvirt-devel`
+Install the libvirt terraform provider:
+```
+GOBIN=~/.terraform.d/plugins go get github.com/crawford/terraform-provider-libvirt
```
enxebre: why is this pointing to a fork? could we just install with tf init?

hack/prebuild/local-dev/main.tf
```diff
@@ -0,0 +1,29 @@
...
+    }]
+  }
+
+  autostart = true
+}
+
+resource "libvirt_volume" "rh_cos_base" {
+  name   = "rh_cos_base"
+  source = "${var.libvirt_base_image}"
+}
```
enxebre: we'll need:
- A default pool https://github.com/openshift/installer/blob/master/docs/dev/libvirt-howto.md#configure-default-libvirt-storage-pool
- The base volume
- Ignition  config
See https://github.com/openshift/machine-api-operator/blob/master/machines/libvirt/worker.machineset.yaml#L29

hack/prebuild/local-dev/variables.tf
```diff
@@ -0,0 +1,46 @@
+variable "tectonic_base_domain" {
```
enxebre: why do we need this?

hack/prebuild/local-dev/main.tf
```diff
@@ -0,0 +1,29 @@
+provider "libvirt" {
+  uri = "${var.libvirt_uri}"
+}
+
+resource "libvirt_network" "dev_net" {
+  name   = "${var.libvirt_network_name}"
+  mode   = "nat"
+  bridge = "${var.libvirt_network_if}"
+  domain = "${var.base_domain}"
```
enxebre: does this variable exist?

enxebre:
we need:
- A default pool https://github.com/openshift/installer/blob/master/docs/dev/libvirt-howto.md#configure-default-libvirt-storage-pool
- The base volume
- Ignition config
See https://github.com/openshift/machine-api-operator/blob/master/machines/libvirt/worker.machineset.yaml#L29

TF seems ok abstraction to me here since we are replicating installer assumptions (which uses tf as well) to me and the actuator reuses tf plugin implementation, however I'm ok with having any other approach if we feel more comfortable with it, as far as we have something in place for testing basic operations with the actuator. This will be evolving quickly, once all becomes more stable we might wanna tests against installer/vanilla cluster so we test more complex features like draining nodes on delete or similar

hack/prebuild/local-dev/main.tf
trawler: @enxebre seems like the terraform libvirt plugin does not manage storage pools (https://github.com/dmacvicar/terraform-provider-libvirt/issues/76). I suggest to use a combination of bash/terraform. What do you think?

paulfantom: How would that look like? Terraform running bash script to provision volumes?

trawler: @paulfantom I'm not a fan of this solution either... but technically, yes, it can be done, if we just want to run a simple `terraform apply` in one go and not add one manual step.

hack/prebuild/local-dev/README.md
trawler: @enxebre at the moment, terraform can only automatically install plugins officially distributed by hashicorp, so we need to manually install it: https://www.terraform.io/docs/configuration/providers.html#third-party-plugins

hack/prebuild/local-dev/main.tf
trawler: both approaches are valid. let's see if anyone else has any preference, one way or another.

paulfantom:
If you are creating a bash script to do provisioning on local machine by using `local-exec` then whole terraform starts to make no sense to me.

It starts to be more like automation for the sake of automation. Doesn't we just need to download rhcos image, create volumes and ignition config? We can use default networking and default storage pool if it was just written in readme.

hack/prebuild/local-dev/variables.tf
```diff
@@ -0,0 +1,61 @@
+variable "base_domain" {
+  type        = "string"
+  description = "The base DNS domain of the cluster."
+  default     = "tt.testing"
+}
+
+variable "libvirt_uri" {
+  type        = "string"
+  description = "libvirt connection URI"
+  default     = "qemu+tcp://192.168.122.1/system"
```
paulfantom: Default connection for libvirt is `qemu:///system`. Why are we using a different one which needs first to be allowed in libvirt config?

trawler: This is the default libvirt network, at least in my virbr0. Same value is used on the installer.
I'll switch it to `qemu:///system`.

hack/prebuild/local-dev/README.md
```diff
@@ -0,0 +1,44 @@
...
+_(optional, but makes subsequent runs a bit faster)_
+```
+cat <<EOF > $HOME/.terraformrc
+plugin_cache_dir = "$HOME/.terraform.d/plugin-cache"
+EOF
+```
+
+## Run terraform
+```
+> terraform init
```
frobware: Drop the `>` characters. The previous `cat` command doesn't have one and, in many ways, it is easier without them as you can triple-click, copy/paste directly into your $SHELL.

PR #35, opened 2018-10-04
use ubuntu in packet.net

paulfantom:
/hold

/assign @ingvagabund @enxebre


paulfantom:
Not needed due to #39

PR #37, opened 2018-10-04
Populate machine status with libvirt domain details

bison:
This populates the machine's status subresource with the libvirt domain's details.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,4 +30,4 @@ RUN GOPATH=/go CGO_ENABLED=0 GOOS=linux go install -a -ldflags '-extldflags "-st
 FROM openshift/origin-base
 RUN yum install -y ca-certificates libvirt-libs openssh-clients
 
-COPY --from=builder /go/bin/machine-controller /go/bin/controller-manager .
+COPY --from=builder /go/bin/machine-controller /go/bin/controller-manager /
--- a/cloud/libvirt/actuators/machine/actuator.go
+++ b/cloud/libvirt/actuators/machine/actuator.go
@@ -14,18 +14,41 @@
 package machine
 
 import (
+	"bytes"
 	"fmt"
 
 	"github.com/golang/glog"
-	libvirtutils "github.com/openshift/cluster-api-provider-libvirt/cloud/libvirt/actuators/machine/utils"
 
+	libvirt "github.com/libvirt/libvirt-go"
+
+	libvirtutils "github.com/openshift/cluster-api-provider-libvirt/cloud/libvirt/actuators/machine/utils"
 	providerconfigv1 "github.com/openshift/cluster-api-provider-libvirt/cloud/libvirt/providerconfig/v1alpha1"
 
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	jsonserializer "k8s.io/apimachinery/pkg/runtime/serializer/json"
 	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 	clusterclient "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset"
 )
+
+type errorWrapper struct {
+	cluster *clusterv1.Cluster
+	machine *clusterv1.Machine
+}
+
+func (e *errorWrapper) Error(err error, message string) error {
+	return fmt.Errorf("%s/%s: %s: %v", e.cluster.Name, e.machine.Name, message, err)
+}
+
+func (e *errorWrapper) WithLog(err error, message string) error {
+	wrapped := e.Error(err, message)
+	glog.Error(wrapped)
+	return wrapped
+}
 
 // Actuator is responsible for performing machine reconciliation
 type Actuator struct {
@@ -52,13 +75,22 @@ func NewActuator(params ActuatorParams) (*Actuator, error) {
 // Create creates a machine and is invoked by the Machine Controller
 func (a *Actuator) Create(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
 	glog.Infof("Creating machine %q for cluster %q.", machine.Name, cluster.Name)
+	errWrapper := errorWrapper{cluster: cluster, machine: machine}
+
 	// TODO: hack to increase IPs. Build proper logic in setNetworkInterfaces method
 	a.cidrOffset++
 
-	if err := createVolumeAndDomain(machine, a.cidrOffset, a.kubeClient); err != nil {
-		glog.Errorf("Could not create libvirt machine: %v", err)
-		return fmt.Errorf("error creating machine %v", err)
+	dom, err := createVolumeAndDomain(machine, a.cidrOffset, a.kubeClient)
+	if err != nil {
+		return errWrapper.WithLog(err, "error creating libvirt machine")
 	}
+
+	defer dom.Free()
+
+	if err := a.updateStatus(machine, dom); err != nil {
+		return errWrapper.WithLog(err, "error updating machine status")
+	}
+
 	return nil
 }
 
@@ -71,7 +103,32 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 // Update updates a machine and is invoked by the Machine Controller
 func (a *Actuator) Update(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
 	glog.Infof("Updating machine %v for cluster %v.", machine.Name, cluster.Name)
-	return fmt.Errorf("TODO: Not yet implemented")
+	errWrapper := errorWrapper{cluster: cluster, machine: machine}
+
+	machineProviderConfig, err := machineProviderConfigFromClusterAPIMachineSpec(&machine.Spec)
+	if err != nil {
+		return errWrapper.WithLog(err, "error getting machineProviderConfig from spec")
+	}
+
+	client, err := libvirtutils.BuildClient(machineProviderConfig.URI)
+	if err != nil {
+		return errWrapper.WithLog(err, "failed to build libvirt client")
+	}
+
+	defer client.Close()
+
+	dom, err := libvirtutils.LookupDomainByName(machine.Name, client)
+	if err != nil {
+		return errWrapper.WithLog(err, "failed to look up domain by name")
+	}
+
+	defer dom.Free()
+
+	if err := a.updateStatus(machine, dom); err != nil {
+		return errWrapper.WithLog(err, "error updating machine status")
+	}
+
+	return nil
 }
 
 // Exists test for the existance of a machine and is invoked by the Machine Controller
@@ -92,18 +149,21 @@ func (a *Actuator) Exists(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 	return libvirtutils.DomainExists(machine.Name, client)
 }
 
-// CreateVolumeAndMachine creates a volume and domain which consumes the former one
-func createVolumeAndDomain(machine *clusterv1.Machine, offset int, kubeClient kubernetes.Interface) error {
+// CreateVolumeAndMachine creates a volume and domain which consumes the former one.
+// Note: Upon success a pointer to the created domain is returned.  It
+// is the caller's responsiblity to free this.
+func createVolumeAndDomain(machine *clusterv1.Machine, offset int, kubeClient kubernetes.Interface) (*libvirt.Domain, error) {
 	// decode config
 	machineProviderConfig, err := machineProviderConfigFromClusterAPIMachineSpec(&machine.Spec)
 	if err != nil {
-		return fmt.Errorf("error getting machineProviderConfig from spec: %v", err)
+		return nil, fmt.Errorf("error getting machineProviderConfig from spec: %v", err)
 	}
 
-	// build libvirtd client
+	// TODO: In the current scheme, the client should probably be a
+	// parameter so the caller can free the client connection.
 	client, err := libvirtutils.BuildClient(machineProviderConfig.URI)
 	if err != nil {
-		return fmt.Errorf("failed to build libvirt client: %v", err)
+		return nil, fmt.Errorf("failed to build libvirt client: %v", err)
 	}
 
 	// TODO(alberto) create struct and function for converting machineconfig->libvirtConfig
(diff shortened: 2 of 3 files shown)


bison:
/hold

Still a WIP. Almost there.

cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -62,11 +92,125 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 // Update updates a machine and is invoked by the Machine Controller
 func (a *Actuator) Update(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
 	glog.Infof("Updating machine %v for cluster %v.", machine.Name, cluster.Name)
-	return fmt.Errorf("TODO: Not yet implemented")
+
+	// TODO: Should we lookup by name if this fails?
+	libvirtStatus, err := libvirtutils.ProviderStatus(machine)
+	if err != nil {
+		glog.Errorf("Could not get provider status for update: %v", err)
+		return err
```
frobware: Why do we, in general, log errors, then return an err? Isn't the caller going to log the error, and perhaps with additional context?

cloud/libvirt/actuators/machine/utils/domain.go
```diff
@@ -745,3 +754,125 @@ func DomainExists(machine *clusterv1.Machine) (bool, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to build libvirt client: %s", err)
+	}
+
+	virConn := client.libvirt
+	if virConn == nil {
+		return nil, fmt.Errorf(LibVirtConIsNil)
+	}
+
+	dom, err := virConn.LookupDomainByUUID(uuid[:])
```
frobware: Having got this far... I'm wondering why we pass a pointer to the uuid? Is it because it's an underlying C/native type?

cloud/libvirt/actuators/machine/utils/domain.go
```diff
@@ -745,3 +754,125 @@ func DomainExists(machine *clusterv1.Machine) (bool, error) {
...
+
+	decoder := providerconfigv1.Codecs.UniversalDecoder(providerconfigv1.SchemeGroupVersion)
+	obj, gvk, err := decoder.Decode([]byte(machine.Status.ProviderStatus.Raw), nil, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	status, ok := obj.(*providerconfigv1.LibvirtMachineProviderStatus)
+	if !ok {
+		return nil, fmt.Errorf("unexpected object: %#v", gvk)
```
frobware: Would expressing `%T` be more useful here? i.e., what type did we get?

cloud/libvirt/actuators/machine/utils/domain.go
```diff
@@ -745,3 +754,125 @@ func DomainExists(machine *clusterv1.Machine) (bool, error) {
...
+		return "Shutdown"
+	case libvirt.DOMAIN_CRASHED:
+		return "Crashed"
+	case libvirt.DOMAIN_PMSUSPENDED:
+		return "Suspended"
+	case libvirt.DOMAIN_SHUTOFF:
+		return "Shutoff"
+	}
+
+	return "Unknown"
```
frobware: Make this the `default` case?

cloud/libvirt/actuators/machine/actuator.go
bison: Mostly doing this in the "top-level" Update / Create / etc. methods because they are the boundary between the actuator and the upstream controller. I've run into cases where the controller didn't log the returned errors which caused a lot of wasted time debugging.

cloud/libvirt/actuators/machine/utils/domain.go
bison: This `gvk` struct will have the group, version, and kind information. Honestly, this bit is just lifted from the AWS actuator, but it seems like a reasonable thing to log in this situation.

cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -44,12 +57,29 @@ func NewActuator(params ActuatorParams) (*Actuator, error) {
...
+	uuid, err := libvirtutils.CreateVolumeAndMachine(machine, a.cidrOffset)
+	if err != nil {
 		glog.Errorf("Could not create libvirt machine: %v", err)
 		return fmt.Errorf("error creating machine %v", err)
 	}
+
+	dom, err := libvirtutils.LookupDomainByUUID(machine, uuid)
+	if err != nil {
+		glog.Errorf("Could not lookup created libvirt machine: %v", err)
+		return fmt.Errorf("error looking up created machine %v", err)
```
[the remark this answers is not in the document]
bison: Fixed.

cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -44,12 +57,29 @@ func NewActuator(params ActuatorParams) (*Actuator, error) {
...
+	}
+
+	defer dom.Free()
+
+	if err := a.updateStatus(machine, dom); err != nil {
+		glog.Errorf("Could not update machine status: %v", err)
+		return fmt.Errorf("error updating machine status: %v", err)
```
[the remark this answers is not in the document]
bison: Fixed.

cloud/libvirt/actuators/machine/utils/domain.go
bison: Sure. Done.

bison:
I think this is ready.

cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -62,11 +105,139 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
-	return fmt.Errorf("TODO: Not yet implemented")
+	error := errorWrapper{cluster: cluster, machine: machine}
+
+	// TODO: Should we lookup by name if this fails?
+	libvirtStatus, err := ProviderStatusFromMachine(machine)
+	if err != nil {
+		return error.WithLog(err, "error getting provider status")
+	}
+
+	uuid, err := uuid.Parse(*libvirtStatus.InstanceID)
```
ingvagabund: This panics when `libvirtStatus.InstanceID` is `nil`. In case you call the `Update` from scratch without populating machine status.

ingvagabund: It should be fine to look based on domain name (to be define under `cloud/libvirt/actuators/machine/utils/domain.go`):
```go
func LookupDomainByName(machine *clusterv1.Machine) (*libvirt.Domain, error) {
	machineProviderConfig, err := MachineProviderConfigFromClusterAPIMachineSpec(&machine.Spec)
	client, err := buildClient(machineProviderConfig.URI)
	if err != nil {
		return nil, fmt.Errorf("Failed to build libvirt client: %s", err)
	}

	virConn := client.libvirt
	if virConn == nil {
		return nil, fmt.Errorf(LibVirtConIsNil)
	}

	dom, err := virConn.LookupDomainByName(machine.Name)
	if err != nil {
		return nil, fmt.Errorf("Failed to lookup domain by name: %v", err)
	}

	return dom, nil
}
```

cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -62,11 +105,139 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 // Update updates a machine and is invoked by the Machine Controller
 func (a *Actuator) Update(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
 	glog.Infof("Updating machine %v for cluster %v.", machine.Name, cluster.Name)
-	return fmt.Errorf("TODO: Not yet implemented")
+	error := errorWrapper{cluster: cluster, machine: machine}
```
ingvagabund: Maybe `errorWr`? I got confused by the `error.WithLog`. The `error` is allocated for `error` type. We should not override it unless necessary.

Dockerfile
```diff
@@ -15,6 +15,8 @@
 # Reproducible builder image
 FROM openshift/origin-release:golang-1.10 as builder
 
+RUN yum install -y libvirt-devel
```
paulfantom: Do not add this as `libvirt-devel` is already in `openshift/origin-release:golang-1.10` image

cloud/libvirt/actuators/machine/utils/domain.go
```diff
@@ -757,3 +766,104 @@ func DomainExists(machine *clusterv1.Machine) (bool, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to lookup domain by UUID: %v", err)
+	}
+
+	return dom, nil
+}
+
+// UpdateProviderStatus updates the provider status in-place with info
+// from the given libvirt domain.
+func UpdateProviderStatus(status *providerconfigv1.LibvirtMachineProviderStatus, dom *libvirt.Domain) error {
```
enxebre: any machine api specific logic does not belong here, we can address it after merging and rebasing https://github.com/openshift/cluster-api-provider-libvirt/pull/38

cloud/libvirt/actuators/machine/utils/domain.go
```diff
@@ -551,35 +554,35 @@ func domainDefInit(domainDef *libvirtxml.Domain, name string, machineConfig prov
 	return nil
 }
 
-func createDomain(name string, machineProviderConfig *providerconfigv1.LibvirtMachineProviderConfig, offset int) error {
+func createDomain(name string, machineProviderConfig *providerconfigv1.LibvirtMachineProviderConfig, offset int) (*uuid.UUID, error) {
 	if name == "" {
-		return fmt.Errorf("Failed to create domain, name is empty")
+		return nil, fmt.Errorf("Failed to create domain, name is empty")
```
frobware: I don't think this is 'failed to create domain'. Rather, 'name is empty'. :-/

cloud/libvirt/actuators/machine/utils/domain.go
```diff
@@ -551,35 +554,35 @@ func domainDefInit(domainDef *libvirtxml.Domain, name string, machineConfig prov
...
 	domainDef, err := newDomainDefForConnection(virConn)
 	if err != nil {
-		return fmt.Errorf("Failed to newDomainDefForConnection: %s", err)
+		return nil, fmt.Errorf("Failed to newDomainDefForConnection: %s", err)
 	}
 
 	// Get values from machineProviderConfig
 	if err := domainDefInit(&domainDef, name, *machineProviderConfig); err != nil {
-		return fmt.Errorf("Failed to init domain definition from machineProviderConfig: %v", err)
+		return nil, fmt.Errorf("Failed to init domain definition from machineProviderConfig: %v", err)
```
frobware: We should consistently use `%v` for error. In the few lines above we use `%s`.

cloud/libvirt/actuators/machine/utils/domain.go
```diff
@@ -551,35 +554,35 @@ func domainDefInit(domainDef *libvirtxml.Domain, name string, machineConfig prov
...
 
 	log.Printf("[INFO] setCoreOSIgnition")
 	if machineProviderConfig.IgnKey != "" {
 		if err := setCoreOSIgnition(&domainDef, machineProviderConfig.IgnKey); err != nil {
-			return err
+			return nil, err
 		}
 	} else {
-		return fmt.Errorf("machine does not has a IgnKey value")
+		return nil, fmt.Errorf("machine does not has a IgnKey value")
```
frobware: typo: `machine does not _have_ a IgnKey value`.

cloud/libvirt/actuators/machine/utils/domain.go
bison: Ended up dropping the UUID thing altogether and just looking up by name. It's simpler and should be fine.

cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -44,12 +72,27 @@ func NewActuator(params ActuatorParams) (*Actuator, error) {
...
+
 	// TODO: hack to increase IPs. Build proper logic in setNetworkInterfaces method
 	a.cidrOffset++
-	if err := libvirtutils.CreateVolumeAndMachine(machine, a.cidrOffset); err != nil {
-		glog.Errorf("Could not create libvirt machine: %v", err)
-		return fmt.Errorf("error creating machine %v", err)
+
+	uuid, err := libvirtutils.CreateVolumeAndMachine(machine, a.cidrOffset)
+	if err != nil {
+		return error.WithLog(err, "error creating libvirt machine")
```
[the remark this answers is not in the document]
bison: Sure, but that's upstream and this means we can actually debug now.

Anyway, this was logging the error via the wrapper. The naming is just confusing. I changed it as suggested in another comment.

cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -62,11 +105,139 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+
+	return err
+}
+
+// EncodeLibvirtMachineProviderStatus encodes a libvirt provider
+// status as a runtime.RawExtension for inclusion in a MachineStatus
+// object.
+func EncodeLibvirtMachineProviderStatus(status *providerconfigv1.LibvirtMachineProviderStatus) (*runtime.RawExtension, error) {
+	// TODO: Modifying this in place feels weird. Should it be copied?
+	status.TypeMeta = metav1.TypeMeta{
```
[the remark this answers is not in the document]
bison: This is what the AWS actuator does, but yeah, it felt weird. I don't think it's needed if we just set the metadata upfront when we first create a status, so I've done that instead and removed this bit.

cloud/libvirt/actuators/machine/actuator.go
bison: I ditched the whole UUID thing. It didn't make sense to look up the domain via a field stored in the status in order to update the status anyway. The domain is looked up by name now.

cloud/libvirt/actuators/machine/actuator.go
bison: Good point. Fixed.

Dockerfile
bison: Looks like this was already fixed in master. This bit is gone after a rebase.

cloud/libvirt/actuators/machine/utils/domain.go
bison: Fixed.

bison:
Rebased and simplified this quite a bit. I think it's probably ready as-is, but a few open questions:

1. The status field names are copied from the AWS actuator. Should we update them to make more sense here? e.g. `DomainUUID` instead of `InstanceID`, or is there a compelling reason to have them match the AWS status?

2. I haven't included any conditions in the status. Happy to do that now if we think it's important, or can tackle it in a follow up PR.

frobware:
> 2\. I haven't included any conditions in the status. Happy to do that now if we think it's important, or can tackle it in a follow up PR.

I vote w do that in a follow-up PR. Better to keep them small and easily reviewable (and revertable).

PR #40, opened 2018-10-06
Initialize rand.Seed to generate random MAC addresses

ingvagabund:
Once build, the actuator always generate the same MAC address for all machines.

Thus, the seed needs to be set to a value that produces different random sequence no matter when the MAC address is generated.

Based on https://stackoverflow.com/questions/12321133/golang-random-number-generator-how-to-seed-properly

### DIFF
--- a/cloud/libvirt/actuators/machine/utils/domain.go
+++ b/cloud/libvirt/actuators/machine/utils/domain.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"os"
 	"strings"
+	"time"
 
 	"math/rand"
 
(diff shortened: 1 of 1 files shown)


cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -35,6 +37,7 @@ type ActuatorParams struct {
 
 // NewActuator creates a new Actuator
 func NewActuator(params ActuatorParams) (*Actuator, error) {
+	rand.Seed(time.Now().UTC().UnixNano())
```
frobware: You could do this in a `init()` function.

enxebre: nice, thanks. Can we include a comment to include context for others when seeing this?

ingvagabund: Ha, that's even better :) I will update the PR.

ingvagabund:
@frobware @enxebre moved the `rand.Seed` under `cloud/libvirt/actuators/machine/utils/domain.go` so it is in the same file as the `rand.Read`.

PR #41, opened 2018-10-08
Reproducible build system

paulfantom:
- Port Makefile implementation from machine-api-operator
- remove additional Makefile and move it's core functions to main one
- Remove installation of libvirt devel libraries from Dockerfile as they are shipped with builder image.

Unit tests will fail as they try to run containerized versions. This will be fixed in openshift/release repo.


paulfantom:
@enxebre fixed. Also added entry to .gitignore to prevent this from happening in the future.

Makefile
```diff
@@ -12,10 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+#REGISTRY    ?= quay.io/openshift/
```
frobware: If we don't need this, then let's drop it.

enxebre:
is there a counterpart PR on release repo for the unit tests?

paulfantom:
Yes, it is here: https://github.com/openshift/release/pull/1840

paulfantom:
@enxebre @frobware openshift/release#1840 needs to be merged before this PR can be auto-merged.

PR #43, opened 2018-10-08
Use faster packet instance

paulfantom:
use `c1.small.x86` (named `baremetal_1`) instead of `t1.small.x86` (`baremetal_0`)

/assign @ingvagabund


ingvagabund:
/hold
let's wait until I test how much time it takes to provision entire cluster API stack with the current setting

PR #52, opened 2018-10-10
minor cleanup

paulfantom:
- dump packet instance ip to a file
- decouple provisioning in makefile
- remove boilerplate.go.txt file

This should allow to run e2e test suite and also allow to do only a provisioning by executing `make e2e-provision`


Makefile
```diff
@@ -66,11 +66,15 @@ integration: deps-cgo ## Run integration test
...
 .PHONY: e2e
-e2e: ## Run end-to-end test
-	hack/packet-provision.sh install
-	#TODO run tests
+e2e: e2e-provision ## Run end-to-end test
+	# TODO @ingvagabund @spangenberg add e2e test command here
 	hack/packet-provision.sh destroy
 
+.PHONY: e2e-provision
+e2e-provision: /tmp/packet_ip
```
frobware: Why can't this be just: 

    e2e-provision:
            hack/packet-provision.sh install

It's not clear what the intermediate target `/tmp/packet_ip` is for.

hack/packet-provision.sh
```diff
@@ -39,7 +39,7 @@ case ${1} in
     ;;
   "destroy")
     terraform destroy -input=false -auto-approve
-    rm /tmp/packet_id_rsa* 2>/dev/null || :
+    rm /tmp/packet_id_rsa* /tmp/packet_ip 2>/dev/null || :
```
frobware: Use `rm -f`, then you don't need the `|| :`.

hack/packet-provision.sh
```diff
@@ -39,7 +39,7 @@ case ${1} in
     ;;
   "destroy")
     terraform destroy -input=false -auto-approve
-    rm /tmp/packet_id_rsa* /tmp/packet_ip 2>/dev/null || :
+    rm -f /tmp/packet_id_rsa* /tmp/packet_ip 2>/dev/null
```
frobware: Why do we need to hide errors? (i.e., the `2>/dev/null`.

paulfantom: No reason.

paulfantom:
I wonder if this will work :smiling_imp: :

/approve

PR #53, opened 2018-10-10
add @bison as a repo owner

paulfantom:
SSIA


bison:
/lgtm

Does my vote count yet...?

PR #55, opened 2018-10-10
Fix leaking libvirt connections

frobware:
We were missing calls to Close() for every call to BuildClient (which
opens a new connection to libvirtd).

Fixes: https://jira.coreos.com/browse/CLOUD-221


frobware:
Closing this as same fix done in https://github.com/openshift/cluster-api-provider-libvirt/pull/37

PR #59, opened 2018-10-11
close libvirt client connection on volume and domain deletion

paulfantom:
SSIA

/assign @bison @frobware

### DIFF
--- a/cloud/libvirt/actuators/machine/actuator.go
+++ b/cloud/libvirt/actuators/machine/actuator.go
@@ -210,6 +210,7 @@ func deleteVolumeAndDomain(machine *clusterv1.Machine) error {
 	if err != nil {
 		return fmt.Errorf("Failed to build libvirt client: %s", err)
 	}
+	defer client.Close()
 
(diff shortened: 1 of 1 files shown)


cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -210,6 +210,7 @@ func deleteVolumeAndDomain(machine *clusterv1.Machine) error {
 	if err != nil {
 		return fmt.Errorf("Failed to build libvirt client: %s", err)
 	}
+	defer client.Close()
```
frobware: Does your change look similar to https://github.com/openshift/cluster-api-provider-libvirt/pull/55

@bison and I discussed this and decided it would be better to set the libvirt connection up in the Create/Update/etc methods, passing the connection in, and potentially sharing it. Here you have to go verify that the call to DeleteDomain doesn't hold on to the client [connection]. And if it doesn't today, it may not be obvious later on when another change is made to this function. It seems better to pass this in as a dependency.

PR #60, opened 2018-10-11
Fix connection leak

bison:
Two quick fixes:

1) Create libvirt client connections in the top-level Create / Update / etc. methods and ensure they are closed. This should fix the leak.

2) Clean up volumes if domain creation fails. Otherwise, subsequent attempts to create the volume and domain will fail because the volume already exists.

### DIFF
--- a/cloud/libvirt/actuators/machine/actuator.go
+++ b/cloud/libvirt/actuators/machine/actuator.go
@@ -77,10 +77,17 @@ func (a *Actuator) Create(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 	glog.Infof("Creating machine %q for cluster %q.", machine.Name, cluster.Name)
 	errWrapper := errorWrapper{cluster: cluster, machine: machine}
 
+	client, err := clientForMachine(machine)
+	if err != nil {
+		return errWrapper.WithLog(err, "error creating libvirt client")
+	}
+
+	defer client.Close()
+
 	// TODO: hack to increase IPs. Build proper logic in setNetworkInterfaces method
 	a.cidrOffset++
 
-	dom, err := createVolumeAndDomain(machine, a.cidrOffset, a.kubeClient)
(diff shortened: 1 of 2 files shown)


cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -197,29 +198,15 @@ func createVolumeAndDomain(machine *clusterv1.Machine, offset int, kubeClient ku
...
-
-	// build libvirtd client
-	client, err := libvirtutils.BuildClient(machineProviderConfig.URI)
-	if err != nil {
-		return fmt.Errorf("Failed to build libvirt client: %s", err)
-	}
-
-	// delete domain
-	if err := libvirtutils.DeleteDomain(name, client); err != nil {
+func deleteVolumeAndDomain(machine *clusterv1.Machine, client *libvirtutils.Client) error {
```
frobware: This function does too much. From the callers perspective we should just `deleteDomain()` and it that fails it can go onto doing other things like `deleteVolume()`. And vice-versa. Trying to do both in one shot has the unexpected side effect of leaving one resource behind. If we're going to continue down this path we should fall- through and attempt to `deleteVolume()` if this call fails.

bison: I don't disagree at all, but seems out of scope here. I really just want to fix the connection leak. I only threw in the fix for the volume cleanup because it was trivial.

frobware: > I only threw in the fix for the volume cleanup because it was trivial.

It was that that made me comment. :)

frobware: Sure, let's keep changes discrete.

frobware:
This certainly fixes a lot of the cases and we grow only slowly at the moment so let's land this as-is and continue to look at the problem in a follow-up PR.

/approve

PR #64, opened 2018-10-15
Makefile: add machine-controller to build target

frobware:
Invoking `make build` now builds both the libvirt-actuator and the
machine-controller.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,12 +43,10 @@ depend:
 depend-update:
 	dep ensure -update
 
-bin:
-	@mkdir $@
-
 .PHONY: build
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -43,12 +43,10 @@ depend:
 depend-update:
 	dep ensure -update
 
-bin:
-	@mkdir $@
-
 .PHONY: build
-build: | bin ## build binary
```
paulfantom: why delete `bin` directory?

frobware: It gets created automatically by `go build -o bin/...`.

PR #68, opened 2018-10-17
Fix machine delete

vikaschoudhary16:
Currently if actuator fails to create a machine and user wants to delete the machine object, kubectl hangs forever. This is because in the reconcile function, error is thrown because of failing to find machine. Kubectl hangs because of the finalizer.

This PR adds logic to first verify if machine exists at the actuator before executing Delete at actuator.

Fixes: https://jira.coreos.com/browse/CLOUD-225

### DIFF
--- a/pkg/cloud/libvirt/actuators/machine/actuator.go
+++ b/pkg/cloud/libvirt/actuators/machine/actuator.go
@@ -121,8 +121,15 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 	}
 
 	defer client.Close()
-
-	return deleteVolumeAndDomain(machine, client)
+	exists, err := libvirtutils.DomainExists(machine.Name, client)
+	if err != nil {
+		return err
+	}
+	if exists {
+		return deleteVolumeAndDomain(machine, client)
+	}
+	glog.Infof("Domain %s does not exist. Skipping deletion...", machine.Name)
+	return nil
 }
 
 // Update updates a machine and is invoked by the Machine Controller


cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -112,8 +112,12 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 	}
 
 	defer client.Close()
-
-	return deleteVolumeAndDomain(machine, client)
+	exists, err := libvirtutils.DomainExists(machine.Name, client)
+	if exists {
+		return deleteVolumeAndDomain(machine, client)
+	} else {
```
frobware: Drop the `else`.

enxebre: also can we add a log saying something like "skipping deletion, machine does not exist" or so

trawler:
This PR doesn't cover situations where the deletion error occurs as a result of connectivity error (when  the workers were not created due to libvirt being down, for example).

pkg/cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -121,8 +121,13 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 	}
 
 	defer client.Close()
-
-	return deleteVolumeAndDomain(machine, client)
+	exists, err := libvirtutils.DomainExists(machine.Name, client)
+	if exists {
+		return deleteVolumeAndDomain(machine, client)
+	} else {
```
enxebre: don't need the `else` here right?

vikaschoudhary16: log message is being printed only in else case, thats the case when we are skipping deletion

vikaschoudhary16:
@trawler i think it does cover. Even in the case you mentioned, same logic will apply because actuator will find that worker does not exist and will not return error.

pkg/cloud/libvirt/actuators/machine/actuator.go
enxebre: yeh, still think don't need the `else` right? It'll only log when `exist` is false 
how about
```
	exists, err := libvirtutils.DomainExists(machine.Name, client)
	if err != nil {
		return err
	}
	if exists {
		return deleteVolumeAndDomain(machine, client)
	}
	glog.Infof("Machine does not exist. Skipping deletion...")
	return nil
```

vikaschoudhary16: i am so dumb. you are right :D

ingvagabund:
@vikaschoudhary16 is the description still valid?

vikaschoudhary16:
@ingvagabund i think so. What you think is invalid?

pkg/cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -121,8 +121,12 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 	}
 
 	defer client.Close()
-
-	return deleteVolumeAndDomain(machine, client)
+	exists, err := libvirtutils.DomainExists(machine.Name, client)
+	if exists {
+		return deleteVolumeAndDomain(machine, client)
+	}
+	glog.Infof("Machine does not exist. Skipping deletion...")
```
ingvagabund: Given the context of the actuator it could make more sense to use `domain` instead of `machine`. E.g.
```go
glog.Infof("Domain %v does not exist. Skipping deletion...", machine.Name)
```

WDYT?

pkg/cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -121,8 +121,12 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 	}
 
 	defer client.Close()
-
-	return deleteVolumeAndDomain(machine, client)
+	exists, err := libvirtutils.DomainExists(machine.Name, client)
```
ingvagabund: This may get evaluated incorrectly in case there are two namespaces with identical machine names. The existence test needs to take the namespace into account as well.

ingvagabund:
> @ingvagabund i think so. What you think is invalid?

Reading for the second time it looks ok. I was confused by the `finalizer` keyword. Now it make sense. Sorry for the noise.

pkg/cloud/libvirt/actuators/machine/actuator.go
ingvagabund: @enxebre looks like we completely ignore the namespace https://github.com/openshift/cluster-api-provider-libvirt/blob/12a25c9b1c1cf88311d518737c7363d260f8b5bc/pkg/cloud/libvirt/actuators/machine/actuator.go#L184

vikaschoudhary16: you mean k8s namespaces? 
Though i dont think it is a valid use case to have machine objects in different namespaces, for example installer creates machines only in openshift-cluster-api namespace. Still if we want to secure this scenario, then we should create domain name using a combination of name and namespace throughout the actuator. I think that should be handled in a seperate PR.

pkg/cloud/libvirt/actuators/machine/actuator.go
vikaschoudhary16: sg

enxebre: this seems still counterintuitive syntax to me, before checking `if exists` we should check the error
```
	exists, err := libvirtutils.DomainExists(machine.Name, client)
	if err != nil {
		return err
	}
	if exists {
		return deleteVolumeAndDomain(machine, client)
	}
	glog.Infof("Machine does not exist. Skipping deletion...")
	return nil
```

pkg/cloud/libvirt/actuators/machine/actuator.go
ingvagabund: That's a valid argument. In that case we should search for a domain based on UUID instead.

vikaschoudhary16: we cant search using UUID because UUID is not used while creating the domain. IMHO, this "uniqueness" should be handled in a later PR.

vikaschoudhary16:
```
./machines.test: /lib64/libvirt.so.0: version `LIBVIRT_4.1.0' not found (required by ./machines.test)
./machines.test: /lib64/libvirt.so.0: version `LIBVIRT_4.4.0' not found (required by ./machines.test)
./machines.test: /lib64/libvirt.so.0: version `LIBVIRT_4.5.0' not found (required by ./machines.test)
++ export status=FAILURE
++ status=FAILURE
```
e2e is failing because of ^, which seems unrelated to changes in the PR.

PR #69, opened 2018-10-17
add tags to instances for easier cleanup

paulfantom:
This adds automatic packet.net instance tagging, just in case we will leak resources.

It uses jenkins node name as tag if run on jenkins and `usermachine` otherwise.

/CC @openshift/sig-cloud

### DIFF
--- a/hack/packet-provision.sh
+++ b/hack/packet-provision.sh
@@ -17,6 +17,7 @@ if [ "$TF_VAR_packet_project_id" == "" ]; then
 fi
 
 export TF_VAR_id=${ID:-$(uuidgen | cut -c1-8)}
+export TF_VAR_tag=${NODE_NAME:-usermachine}
 
 pushd $script_dir/prebuild
 
--- a/hack/packet-provision.sh
+++ b/hack/packet-provision.sh
@@ -17,7 +17,7 @@ if [ "$TF_VAR_packet_project_id" == "" ]; then
 fi
 
 export TF_VAR_id=${ID:-$(uuidgen | cut -c1-8)}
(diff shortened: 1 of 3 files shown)


paulfantom:
I wonder if this works:

/assign @openshift/sig-cloud

hack/packet-provision.sh
```diff
@@ -17,6 +17,7 @@ if [ "$TF_VAR_packet_project_id" == "" ]; then
 fi
 
 export TF_VAR_id=${ID:-$(uuidgen | cut -c1-8)}
+export TF_VAR_tag=${NODE_NAME:-usermachine}
```
frobware: If `NODENAME` is unset (for everyone) then this become `usermachine` (for everyone). Yes? Do we want to add a prefix that help with identity?

paulfantom: Yes, that's the idea. `NODE_NAME` is a variable always set by jenkins. Essentially this is a failsafe aimed at CI system so I wasn't creating anything  for users.

paulfantom:
Fixed issue with wrong variable type and added dynamic tag name if `NODE_NAME` is not present.

hack/prebuild/main.tf
```diff
@@ -10,6 +10,7 @@ resource "packet_device" "libvirt" {
   operating_system = "centos_7"
   billing_cycle    = "hourly"
   project_id       = "${var.packet_project_id}"
+  tags             = "${list("${var.tag}")}"
```
frobware: Why is the `${list...}` now required?

paulfantom: because `tags` needs a list type, not a string

PR #75, opened 2018-10-23
remove copyright notice

paulfantom:
Part of CLOUD-244


ingvagabund:
We should generate diffferent boillerplare header since we need to express the projects resources are under ASL 2.0 license.

paulfantom:
Ideas on a content of such boilerplate? If we just want to say that this is using ASL 2.0, so maybe only SPDX header?

paulfantom:
Also whole repo is ASL 2.0 :)

PR #87, opened 2018-11-05
fix typos and remove ineffectual assignment

paulfantom:
SSIA

### DIFF
--- a/pkg/cloud/libvirt/actuators/machine/utils/domain.go
+++ b/pkg/cloud/libvirt/actuators/machine/utils/domain.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/cloud/libvirt/actuators/machine/utils/domain.go
```diff
@@ -497,9 +497,6 @@ func CreateDomain(name, ignKey, volumeName, hostName, networkInterfaceName, netw
 
 	glog.Infof("setDisks")
 	VolumeKey := baseVolumePath + volumeName
```
enxebre: could you also update this to use https://golang.org/pkg/path/#Join

pkg/cloud/libvirt/actuators/machine/utils/domain.go
```diff
@@ -497,9 +497,6 @@ func CreateDomain(name, ignKey, volumeName, hostName, networkInterfaceName, netw
...
-	if volumeName == "" {
```
enxebre: we should ideally narrow down the number of parameters injected, encapsulate them into a struct, validate the inputs/error when relevant and unit test it. I understand might be out of the scope for your current work

pkg/cloud/libvirt/actuators/machine/utils/domain.go
paulfantom: Done

pkg/cloud/libvirt/actuators/machine/utils/domain.go
paulfantom: that would be perfect, but it is a bit out of scope

paulfantom:
Will be merged in #92

PR #88, opened 2018-11-05
catch errors and print them

paulfantom:
Some of `err` variables weren't used. This PR catches them and prints messages when they are not `nil`


pkg/cloud/libvirt/actuators/machine/utils/domain.go
```diff
@@ -356,6 +356,9 @@ func setNetworkInterfaces(domainDef *libvirtxml.Domain,
 				return fmt.Errorf("Error retrieving network name: %s", err)
 			}
 			networkDef, err := newDefNetworkfromLibvirt(network)
+			if err != nil {
+				return fmt.Errorf("Error retrieving network definition: %s", err)
```
enxebre: should be %v

PR #90, opened 2018-11-05
WIP: Revert "remove log-level option"

ingvagabund:
Reverts openshift/cluster-api-provider-libvirt#82

Just to see if this is the reason why the e2e fails.


paulfantom:
Just to be sure
/hold

enxebre:
can we just update https://github.com/openshift/cluster-api-provider-libvirt/blob/master/pkg/cloud/libvirt/actuators/machine/actuator.go#L309 to call ```a.clusterClient.ClusterV1alpha1().
        Machines(machineCopy.Namespace).UpdateStatus(machineCopy)``` and update the code to use the client provider by the manager controller as we do for aws

PR #92, opened 2018-11-05
Update packet image

paulfantom:
Image was build and is stored in https://github.com/paulfantom/packet-image

/assign @ingvagabund


enxebre:
we need to:
- revendor the framework to pick the crd status fixes
- Update the client calls to updaStatus
- Update `hack/prebulid/init.sh` to pin to kube version 1.11.3

paulfantom:
> revendor the framework to pick the crd status fixes

Framework is in latest version.

PR #96, opened 2018-11-16
Update machine status correctly

vikaschoudhary16:


frobware:
@vikaschoudhary16 - can we add some description for commit message. SSIA never tells me anything, particularly "why/what" when I `git log` at some later date.

pkg/cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -307,8 +307,8 @@ func (a *Actuator) applyMachineStatus(
 	now := metav1.Now()
 	machineCopy.Status.LastUpdated = &now
 	_, err = a.clusterClient.ClusterV1alpha1().
-		Machines(machineCopy.Namespace).Update(machineCopy)
-
+		Machines(machineCopy.Namespace).UpdateStatus(machineCopy)
+	glog.Infof("error in updating: %v", err)
```
frobware: If there's no error we will print nothing meaningful which will be misleading. Also, if it's an error then `.Errorf`.

frobware:
@vikaschoudhary16 - great commit message. 👍 

/lgtm

PR #99, opened 2018-11-25
Deploy manager controller with machine controller

ingvagabund:
Test before merging it in `actuator-pkg` repo.

/hold


ingvagabund:
Just applying https://github.com/openshift/cluster-api-actuator-pkg. We need to sync cluster-api dep accros all our repos in a single step. @frobware @enxebre

frobware:
As this now looks like just a vendoring change:

/lgtm

PR #104, opened 2018-12-23
Accomodate provider spec

ingvagabund:
Built on top of https://github.com/openshift/cluster-api-provider-libvirt/pull/103

Due to the following breaking change https://github.com/kubernetes-sigs/cluster-api/pull/548 we need to support both `.providerConfig` and `.providerSpec` fields until all consumers of the libvirt actuator migrate to new `.providerSpec` field.

We could just keep the `.providerConfig` and make sure it gets de-serialized into `ProviderSpec` struct. Though, folks could get confused by that and we should stay with the upstream name convention.

Once merged, we can rename the following:
- `providerConfig`
- `ProviderConfig`
- `DecodeProviderConfig`
- `EncodeProviderConfig`
- `LibvirtMachineProviderConfig`

to corresponding:
- `providerSpec`
- `ProviderSpec`
- `DecodeProviderSpec`
- `EncodeProviderSpec`
- `LibvirtMachineProviderSpec`

plus other identifiers that use `provider{c,C}onfig` string.


Gopkg.toml
```diff
@@ -20,10 +20,11 @@ required = [
 
 [[constraint]]
   name = "sigs.k8s.io/cluster-api"
+  #revision = "0734939e05aeb64ab198e3feeee8b4e90ee5cbb2"
   branch = "master"
```
cynepco3hahue: I think it better to use tag `0.0.0-alpha.4` because it includes all recent changes and will prevent broken API on `deps update`

ingvagabund: That release does not contain https://github.com/kubernetes-sigs/cluster-api/pull/598. It's been unwritten rule to take a specific revision or master HEAD til now. Given we usually need the latest changes in the project, taking anything than master HEAD can be problematic.

cynepco3hahue: I see your point but can we at least stick to some commit on the master?

ingvagabund: If we do it here, we need to do it everywhere else. Though, we can wait until the next release is out and update all deps at once in all repos.

cmd/libvirt-actuator/main.go
```diff
@@ -309,6 +309,12 @@ func bootstrapCommand() *cobra.Command {
 				glog.Info(msg)
 			}
 
+			clusterFramework.MachineControllerImage = "openshift/origin-libvirt-machine-controllers:v4.0.0"
+			clusterFramework.MachineManagerImage = "openshift/origin-libvirt-machine-controllers:v4.0.0"
+			// TODO(jchaloup): once docker.io/openshift/origin-machine-api-operator gets updated, switch back to it.
```
paulfantom: It should get an update today, as https://github.com/openshift/machine-api-operator/pull/159 is on its way.

ingvagabund: Who is responsible for making the updates?

paulfantom: prow

paulfantom:
/approve

putting it on hold since we should have a new mao image in docker hub today (this should do it: https://github.com/openshift/machine-api-operator/pull/159) and in turn  new image should resolve a TODO.

/hold

cmd/libvirt-actuator/main.go
ingvagabund: How often is it done?

paulfantom: As far as I remember it should be done every couple of hours

ingvagabund: The latest mao docker images is two weeks old :-|

paulfantom: It seems like like CI image mirroring responsible for uploading images to docker hub was down between ~13 days ago and ~43 minutes ago.

Gopkg.toml
ingvagabund: Otherwise, the exact sha of the cluster-api is in the Godep.lock file.

Makefile
```diff
@@ -83,7 +83,7 @@ build-e2e:
 
 .PHONY: test-e2e
 test-e2e: images build-e2e e2e-provision ## Run end-to-end test
-	hack/test-e2e.sh || $(MAKE) e2e-clean
+	hack/test-e2e.sh || ($(MAKE) e2e-clean && false)
```
frobware: Why `&& false`?

pkg/cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -308,12 +308,22 @@ func (a *Actuator) deleteVolumeAndDomain(machine *clusterv1.Machine, client libv
 // ProviderConfigMachine gets the machine provider config MachineSetSpec from the
 // specified cluster-api MachineSpec.
 func ProviderConfigMachine(codec codec, ms *clusterv1.MachineSpec) (*providerconfigv1.LibvirtMachineProviderConfig, error) {
-	if ms.ProviderConfig.Value == nil {
+	// TODO(jchaloup): Remove providerConfig once all consumers migrate to providerSpec
+	providerSpec := ms.ProviderConfig
+	// providerSpec has higher priority over providerConfig
+	if ms.ProviderSpec.Value != nil || ms.ProviderSpec.ValueFrom != nil {
+		providerSpec = ms.ProviderSpec
+		fmt.Printf("Falling to default providerSpec\n")
```
frobware: Log instead of printing.

pkg/cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -308,12 +308,22 @@ func (a *Actuator) deleteVolumeAndDomain(machine *clusterv1.Machine, client libv
...
+	} else {
+		fmt.Printf("Falling to providerConfig\n")
```
frobware: Log instead of printing.

Makefile
ingvagabund: Since if `hack/test-e2e.sh` fails we need to clean the libvirt resource and fail

frobware: Why doesn't `hack/test-e2e.sh` exit with non-zero if the test fails? If it doesn't we should fix the root cause.

frobware: Oh, I see. If it fails,we clean up but we still want to return non-zero to make.

ingvagabund:
/hold cancel
we can switch to the docker image at any point.

PR #107, opened 2019-01-11
Remove providerconfig

ingvagabund:
- remove any use of the `.providerConfig` and `ProviderConfig`
- rename testing resources to reflect new provider spec name
- rename [Encode|Decode]ProviderConfig to [Encode|Decode]ProviderSpec

To be merged not before https://github.com/openshift/machine-api-operator/pull/173


enxebre:
can you include a separate commit to remove the deprecated field from the vendored cluste-api i.e machine_types.go

PR #109, opened 2019-01-14
Validate machineset before reconciling

ingvagabund:
Even thought the MachineSet type implements Validate() function,
it's not called by default. The validation function is responsible
for making sure every machine set matchLabels selector is matched with
machine template labels. Given the validation is not performed by default,
it is possible to create an invalid machineset that causes the machineset
controller to start creating machine object one by one without any upper
bound. Causing the machine controller to launch as many instances as
there is machine objects.

Carried as downstream patch before kubernetes-sigs/cluster-api#669 gets resolved.


ingvagabund:
Just updating the Makefile so the patch is actually applied when updating the deps.

bison:
/lgtm

At some point, it may be easier to fork the cluster-api repo and update the dependencies to track a branch there. Then we can rebase and cherry-pick into that branch as necessary rather than copy the patches around.

PR #112, opened 2019-02-05
image: add libvirt-devel package to Dockerfile

paulfantom:
This package is installed in image used for okd builds, but it is not present in openshift images.


enxebre:
why do we need this now?

PR #114, opened 2019-02-15
WIP: Drain node

ingvagabund:
Still need to figure out better way how to actually test. Though, we might just move the node draining logic our of actuator and put it into the machine controller under openshift/cluster-api.


ingvagabund:
This is no longer needed as it gets pulled once we sync with the latest openshift/cluster-api.

PR #115, opened 2019-02-15
Remove any usage of cluster object in actuator

ingvagabund:
Cluster object is no longer needed and is optional.

Currently breaking libvirt deployment: https://github.com/openshift/installer/pull/1249#issuecomment-464138413


cgwalters:
I don't know this code...seems like it'd be better to have some sort of special "unset cluster" object rather than drop the logging?

But eh, let's get libvirt working!

ingvagabund:
We already droped mention of the cluster in aws actuator. Instead of cluster object, we can reuse cluster annotation if available.

PR #125, opened 2019-03-15
Report hostname using dhcp provider hostname for a domain

praveenkumar:
In case of libvirt dom.GetHostname(0) will not provide the hostname info since it need qemu-guest-agent installed on the VM to fetch this info https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainGetHostname Since for libvirt case we are providing the hostname using libvirt provided dhcp and assign same to domain name. We can reuse it here also instead adding another package to RHCOS image.

cc @abhinavdahiya @zeenix

### DIFF
--- a/pkg/cloud/libvirt/actuators/machine/actuator.go
+++ b/pkg/cloud/libvirt/actuators/machine/actuator.go
@@ -465,7 +465,11 @@ func NodeAddresses(dom *libvirt.Domain) ([]corev1.NodeAddress, error) {
 		}
 	}
 
-	hostname, err := dom.GetHostname(0)
+	// In case of libvirt dom.GetHostname(0) will not provide the hostname info since it need qemu-guest-agent
+	// installed on the VM to fetch this info https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainGetHostname
+	// Since for libvirt case we are providing the hostname using libvirt provided dhcp and assign same to domain name
+	// We can reuse it here also instead adding another package to RHCOS image.
+	hostname, err := dom.GetName()
 	if err != nil {
 		return addrs, nil
 	}


pkg/cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -465,7 +465,11 @@ func NodeAddresses(dom *libvirt.Domain) ([]corev1.NodeAddress, error) {
 		}
 	}
 
-	hostname, err := dom.GetHostname(0)
+	// In case of libvirt dom.GetHostname(0) will not provide the hostname info since it need qemu-guest-agent
```
zeenix: nit: `libvirt` mention is redundant here.

zeenix: based on the context, i mean.

zeenix:
I tested this via the image you provided (`docker.io/kumarpraveen/capl:csr_issue`) and it at least fixes the TLS errors I have recently been getting.

cynepco3hahue:
IMHO the right solution is to use information from DHCP leases that libvirt provides via `virsh net-dhcp-leases`.

bison:
> IMHO the right solution is to use information from DHCP leases that libvirt provides via virsh net-dhcp-leases.

I agree. Does #128 alone solve the immediate problem? If so, I'd vote for that.

praveenkumar:
> I agree. Does #128 alone solve the immediate problem? If so, I'd vote for that.

@bison @cynepco3hahue So I don't understand why we need to get the hostname from the dhcp lease since we know them already before creating those VM on libvirt? Do you folks thinks that there might be a situation where name which provided by the dhcp and part of the domain can be differ from what we get from the dhcp lease?

zeenix:
> o I don't understand why we need to get the hostname from the dhcp lease since we know them already before creating those VM on libvirt?

Because it's more future-proof as we don't assume a direct relationship between hostname and name of the VM.

praveenkumar:
closing this in favour of #128

PR #135, opened 2019-03-20
Vendor latest cluster-api changes

cynepco3hahue:
Some interesting latest changes:
- drain the node before instance deletion
- remove the relevant node after the machine deletion


ingvagabund:
@cynepco3hahue please re-sync the cluster-api-actuator-pkg with the latest one more time. We need https://github.com/openshift/cluster-api-actuator-pkg/pull/54 to be included as well.

Issue #28, opened 2018-10-01
support different sizes for worker disks.


dustymabe:
Some people try to resize the provided RH CoreOS image using `qemu-img resize` and then are surprised when workers don't show up with larger disks (see https://github.com/openshift/os/issues/318#issuecomment-424890840). 

Can we implement this?

https://github.com/openshift/cluster-api-provider-libvirt/blob/83c10e8213498c31adc3667eb4457fa6f82bb103/cloud/libvirt/actuators/machine/utils/volume.go#L14-L17

derekwaynecarr:
@dustymabe agree.

PR #138, opened 2019-03-21
Support different sizes for worker disks

michaelgugino:
This commit will ensure newly provisioned instance volume
will utilize the larger of defaultSize or the
base volume capacity.

Fixes: https://jira.coreos.com/browse/CLOUD-202
Fixes: https://github.com/openshift/cluster-api-provider-libvirt/issues/28

### DIFF
--- a/pkg/cloud/libvirt/client/client.go
+++ b/pkg/cloud/libvirt/client/client.go
@@ -379,10 +379,20 @@ func (client *libvirtClient) CreateVolume(input CreateVolumeInput) error {
 		volumeDef.Capacity.Value = size
 	} else if input.BaseVolumeID != "" {
 		volume = nil
-		volumeDef.Capacity.Value = uint64(defaultSize)
+
 		baseVolume, err := client.connection.LookupStorageVolByKey(input.BaseVolumeID)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/cloud/libvirt/client/client.go
```diff
@@ -379,11 +379,21 @@ func (client *libvirtClient) CreateVolume(input CreateVolumeInput) error {
...
+		var baseVolumeInfo *libvirt.StorageVolInfo
+		baseVolumeInfo, err = baseVolume.GetInfo()
+		if err != nil {
+			return fmt.Errorf("Can't retrieve volume info %s", input.BaseVolumeID)
+		}
+		if baseVolumeInfo.Capacity > uint64(defaultSize) {
+			volumeDef.Capacity.Value = baseVolumeInfo.Capacity
+		} else {
+			volumeDef.Capacity.Value = uint64(defaultSize)
+		}
```
ingvagabund: Definitely an improvement for cases where `baseVolumeInfo.Capacity > uint64(defaultSize)`. Guest can consume more space in case he needs volume \in <defaultSize, baseVolumeInfo.Capacity>. Which was not allowed until now.

PR #139, opened 2019-03-27
Sync cluster-api to openshift-4.0-cluster-api-0.0.0-alpha.4

ingvagabund:
SSIA


Gopkg.toml
```diff
@@ -19,9 +19,9 @@ required = [
 [prune]
   go-tests = true
 
-[[constraint]]
+[[override]]
   name = "github.com/openshift/cluster-api"
-  revision = "8b6a416a1830212712a6b03c5431afbd4f586d62"
+  branch = "openshift-4.0-cluster-api-0.0.0-alpha.4"
```
michaelgugino: I have a PR open that conflicts with this.  I think it's better pattern to sync with one dependency instead of two, cluster-api-actuator-pkg.

I'm not sure how override differs from constraint though, so maybe I'm missing something.

ingvagabund:
/hold
until machine-api-operator and aws actuator are bump to k8s 1.13.1

Gopkg.toml
ingvagabund: In case there is a dependency that constrains `github.com/openshift/cluster-api` to a certain revision/branch, `override` cancels that constraint and replaces it with the new one.

I have not updated `cluster-api-actuator-pkg` yet. Still on my TODO list. Waiting until my sync PRs get merged for all actuators and machine-api-operator. Once done, I will update actuator-pkg as well. Then we can remove my constraint and proceed with your PR. Would that be OK?

PR #140, opened 2019-03-27
Use host-passthrough CPU mode for wokers

### LINKED ISSUE #105: Support 'host-passthrough' CPU type (and generally, CPU type) for worker nodes
Currently, the VMs are created with the default CPU type of libvirt. If we could define the CPU type, it'd be great - it'll allow us to use 'host-passthrough' CPU type.

cynepco3hahue:
It makes possible to run KubeVirt workflow on top of the libvirt provider.

Fixes #105

### DIFF
--- a/pkg/cloud/libvirt/client/domain.go
+++ b/pkg/cloud/libvirt/client/domain.go
@@ -426,6 +426,8 @@ func domainDefInit(domainDef *libvirtxml.Domain, name string, memory, vcpu int) 
 		return fmt.Errorf("machine does not have an DomainVcpu set")
 	}
 
+	domainDef.CPU.Mode = "host-passthrough"
+
 	//setConsoles(d, &domainDef)
 	//setCmdlineArgs(d, &domainDef)
 	//setFirmware(d, &domainDef)


ingvagabund:
@mykaul @tareqalayan @fabiand @pavolbrilla @zeenix @fabiand can you guys review the PR? Does it solve or at least does not it break your deployments?

pkg/cloud/libvirt/client/domain.go
```diff
@@ -426,6 +426,8 @@ func domainDefInit(domainDef *libvirtxml.Domain, name string, memory, vcpu int)
 		return fmt.Errorf("machine does not have an DomainVcpu set")
 	}
 
+	domainDef.CPU.Mode = "host-passthrough"
```
fabiand: host-model?

mykaul: I personally prefer host-passthrough. You don't plan migrating them, and host-passthrough is the most compatible one. The only downside is that it works well with KVM, may have less success with QEMU (but why would we need QEMU only virtualization?)

fabiand: I don't mind much.

pavolbrilla: fine by me

zeenix: As i said in the issue, `host-passthrough` is pretty good but if you want the most reliable solution, you need to do a bit of extra work: Ask libvirt for host capabilities and set a custom cpu mode based on the best capability you see. Anyway, no objection on this change.

cynepco3hahue:
@ingvagabund Can we merge this one as well?

zeenix:
@cynepco3hahue Seems a testcase is failing.

cynepco3hahue:
@zeenix nope it is a new line that checks if `cluster-api-actuator-pkg` vendored with latest changes. I do not sure how can I fix it.

ingvagabund:
Assuming reviews are find with the PR:

/approve
/lgtm

PR #45, opened 2018-10-08
drop volumes full paths

enxebre:
Drop full paths in favour of pool and name based look up for volumes
See https://github.com/openshift/machine-api-operator/pull/70

PR #144, opened 2019-04-16
Drop volumes full paths

zeenix:
The path of the storage volume is and should be determined by the containing storage pool and hence we should not associate and assume the path of the volume.

Without this change, it is not possible for user to specify alternative location for storage volumes to Openshift Installer.

**WARNING:** 

 * This will also require a change in the Openshift Installer as that still passess the volume path instead of name to the actuator.

 * I've not testing this yet, only built successfully.

----
This is a revision/rebase of https://github.com/openshift/cluster-api-provider-libvirt/pull/45

### DIFF
--- a/pkg/cloud/libvirt/client/client.go
+++ b/pkg/cloud/libvirt/client/client.go
@@ -36,9 +36,6 @@ type CreateDomainInput struct {
 	// IgnitionVolumeName of ignition volume to be added to domain definition
 	IgnitionVolumeName string
 
-	// VolumePoolName of pool where VolumeName volume is located
-	VolumePoolName string
-
 	// NetworkInterfaceName as name of network interface
 	NetworkInterfaceName string
 
@@ -72,11 +69,8 @@ type CreateVolumeInput struct {
 	// VolumeName to be created
 	VolumeName string
 
-	// PoolName where VolumeName volume is located
-	PoolName string
-
-	// BaseVolumeID as base volume ID
-	BaseVolumeID string
+	// BaseVolumeName as name of the base volume
+	BaseVolumeName string
 
 	// Source as location of base volume
 	Source string
@@ -86,7 +80,7 @@ type CreateVolumeInput struct {
 }
 
 // LibvirtClientBuilderFuncType is function type for building aws client
-type LibvirtClientBuilderFuncType func(URI string) (Client, error)
+type LibvirtClientBuilderFuncType func(URI string, poolName string) (Client, error)
 
 // Client is a wrapper object for actual libvirt library to allow for easier testing.
 type Client interface {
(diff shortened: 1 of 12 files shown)


zeenix:
Hmm.. i see a bunch of issues. I'll fix them!

pkg/cloud/libvirt/client/client.go
```diff
@@ -120,21 +114,33 @@ type Client interface {
 
 type libvirtClient struct {
 	connection *libvirt.Connect
+
+	// storage pool that holds all volumes
+	pool *libvirt.StoragePool
+	// cache pool's name so we don't have to call failable GetName() method on pool all the time.
+	poolName string
```
zeenix: @bison Hi, any opinion on this change: making the storage pool part of the client so we don't need to fetch it all the time and pass the pool or name to all the volume handling functions everywhere? I first thought it'd be a good idea and hence the change here but now I'm not so sure.

zeenix:
The e2e tests won't work I think cause this requires a small change in Installer as well: https://github.com/openshift/installer/pull/1628

zeenix:
> The e2e tests won't work I think cause this requires a small change in Installer as well: [openshift/installer#1628](https://github.com/openshift/installer/pull/1628)

Actually I made the changes backwards compat so existing API should work and hence e2e tests should succeed now. I'll check tomorrow why that's not the case.

ingvagabund:
Repeating error:
```
Error: packet_device.libvirt: "facilities": required field is not set

Error: packet_device.libvirt: "facility": [REMOVED] Use the "facilities" array instead, i.e. change 
  facility = "ewr1"
to 
  facilities = ["ewr1"]
```

ingvagabund:
/approve

no objections here, just waiting until CI goes green

ingvagabund:
@zeenix is the change compatible with the installer? Anything else left to do before the PR merges?

pkg/cloud/libvirt/client/client.go
frobware: > I first thought it'd be a good idea and hence the change here but now I'm not so sure

Could you expand on that? 

Otherwise this /lgtm but just wanted to understand why you mentioned that you were not so sure now.

zeenix: I wasn't sure if there is a 1-1 relationship between `libvirtClient` and storage pools in terms of lifecycle. I asked @ingvagabund on slack yesterday and they said it's ok.

frobware:
> @zeenix is the change compatible with the installer? Anything else left to do before the PR merges?

I'm assuming this needs to merge first per the comments in https://github.com/openshift/installer/pull/1628#issuecomment-485838134

zeenix:
> @zeenix is the change compatible with the installer? Anything else left to do before the PR merges?

I did my best to ensure that it is but I failed to test it since it turns out it's not easy to make Installer launch the cluster with custom-built actuator.

zeenix:
> I'm assuming this needs to merge first per the comments in [openshift/installer#1628 (comment)](https://github.com/openshift/installer/pull/1628#issuecomment-485838134)

Indeed. Thanks for the review and merge.

PR #148, opened 2019-04-24
OWNERS: Add CRC team members

zeenix:
This is to extend the ownership to CRC team.


zeenix:
@ingvagabund The related [Installer PR](https://github.com/openshift/installer/pull/1662) was merged this weekend so I think it'd make sense to merge this as well. Thanks.

PR #153, opened 2019-05-14
Remove workaround for old libvirt bug

cfergeau:
The bug was fixed in libvirt 1.3.3 which was released 3 years ago.

Supported fedora releases ship a much newer version
EL7.3 already had libvirt 2.0.0
Debian stable has libvirt 3.0.0
Ubuntu 18.04 LTS has libvirt 4.0.0


zeenix:
What's with the @openshift-bot continously restarting the tests? The tests are unlikely to succeed. I've never seen them succeed so far at least.

PR #157, opened 2019-06-10
Cache leases IP's and do not use them on the machine creation

### LINKED ISSUE #134: Redeploy of the machine-controller will reset cidrOffset
We increase `cidrOffset` each time that the controller creates a new machine and add the record to the libvirt network, but controller redeploy will reset the `cidrOffset` counter and the controller will create DNS records for the new machine with the IP's that already given to another machine.

I do not sure how can we fix it, but I think we will need to check the correlation between running machines and DHCP leases, it also will help to avoid MAC duplication.

cynepco3hahue:
Workers got IP's from DHCP, by default first machine will get IP x.x.x.51,
the second x.x.x.52... But if machine-controllers deployment restarted, it
will start again from x.x.x.51 for new machines. To avoid it, we will cache
IP leases on the first `Create()` call and will generate new IP until we get
IP that does not exist under the cache.

Fixes #134

### DIFF
--- a/pkg/cloud/libvirt/client/network.go
+++ b/pkg/cloud/libvirt/client/network.go
@@ -4,6 +4,7 @@ import (
 	"encoding/xml"
 	"fmt"
 	"math/rand"
+	"sync"
 
 	"github.com/golang/glog"
 
@@ -16,7 +17,14 @@ const (
 	netModeNat      = "nat"
 	netModeRoute    = "route"
 	netModeBridge   = "bridge"
+	workerIPCidr    = 51
 )
+
+// Leases contains list of DHCP leases
+type Leases struct {
+	Items map[string]string
+	sync.Mutex
+}
 
 // Network interface used to expose a libvirt.Network
 type Network interface {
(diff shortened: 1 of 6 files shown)


zeenix:
No clue what's with the docker image not containing the needed files. Let's try again before we investigate:

/test e2e-libvirt

pkg/cloud/libvirt/client/network.go
```diff
@@ -16,8 +17,15 @@ const (
 	netModeNat      = "nat"
 	netModeRoute    = "route"
 	netModeBridge   = "bridge"
+	workerIPCidr    = 51
 )
 
+// Leases contains list of Lease
```
zeenix: Not the most useful comment. :) `Contains list of DHCP leases` would be better IMO.

zeenix:
other than that lgtm.

pkg/cloud/libvirt/client/network.go
cynepco3hahue: Done

PR #159, opened 2019-06-13
Add Namespace flag

praveenkumar:
This is ported back from https://github.com/openshift/cluster-api-provider-azure/pull/40


zeenix:
/test e2e

@praveenkumar we don't have e2e-libvirt yet here, do we?

PR #161, opened 2019-07-10
Copy over relevant Dockerfiles from Installer repo

### LINKED ISSUE #160: e2e-libvirt CI job broken
The `e2e-libvirt` CI job is currently [not working](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-api-provider-libvirt/157/pull-ci-openshift-cluster-api-provider-libvirt-master-e2e-libvirt/2/build-log.txt) for this provider. The reason is that the job configuration assumes dockerfiles for the images it needs to be present in the repository.

So we either need to somehow instruct the job to look for the docker files in the Installer repository or we need to duplicate the docker files in here. The former would be the best approach but we need to figure out how to do that, the latter is very easy (famous last words) but it's not ideal at all since then we'll need to keep them in sync manually.

zeenix:
The `e2e-libvirt` CI job currently assumes these docker files to be present in the repo it is running against. While it would be great to be able to tell it to get the Dockerfiles from Installer repo, someone needs to figure out how to do that. In the meantime, let's copy over these files to get the CI working.

Fixes #160.


zeenix:
Closed since we realized this won't work, at least not so easily so now looking into creating a different job for cluster-api-provider-libvirt.

zeenix:
I think this is needed after all. https://github.com/openshift/release/pull/4406 fails due to:

```
error: could not run steps: step e2e-libvirt failed: template e2e-libvirt has required parameter LOCAL_IMAGE_LIBVIRT_INSTALLER which is not defined
```

and @sallyom informed me that I need to be building the installer image (with libvirt backend) in the `cluster-api-provider`'s config to fix this. However, adding this to the config:

```
 dockerfile_path: images/nested-libvirt/Dockerfile                                                                                                                                                                
  from: base
  inputs:
    root:
      as:
      - build
      paths: null
  optional: true
  to: libvirt-installer
``` 
alone won't work since this config is working on the cluster API libvirt repository and these dockerfiles are not there. So I'm re-openning this.

zeenix:
@ingvagabund If/when you've a moment, please have a look. I've been banging my head against the super complicated CI system for many days now and could use some help. I'd really love to fix the CI here.

ingvagabund:
@zeenix this goes out of my knowledge unfortunately. @petr-muller might help here?

ingvagabund:
> template e2e-libvirt has required parameter LOCAL_IMAGE_LIBVIRT_INSTALLER which is not defined

@zeenix after talking to petr, it's even out of his powers. Checking, I see @sallyom has made most of the changes in [1]. Not sure who else to ask here.

[1] https://github.com/openshift/release/blame/master/ci-operator/templates/openshift/installer/cluster-launch-installer-libvirt-e2e.yaml

zeenix:
Seems we don't want or even need this after https://github.com/openshift/release/pull/4587 but let's keep it around for testing the `e2e-libvirt`.

/hold

zeenix:
Yay! the `e2e-libvirt` passed. \o/ However, to ensure that the changes here are not needed. I'll push a revert of these changes.

praveenkumar:
@zeenix or we can just close this PR without merge, since there is no changes in the file :)

/close

zeenix:
> @zeenix or we can just close this PR without merge, since there is no changes in the file :)

Ah right, that would go as 'no changes'. Let's create another PR.

PR #167, opened 2019-08-01
DON'T MERGE!!!!

zeenix:
Break the provider on purpose to test the CI job.


zeenix:
yay, it fails as hoped. :) Let's give it one more go before closing this PR.

/test e2e-libvirt

Issue #171, opened 2019-08-23
client.CreateVolume doesn't check the returned error of waitForSuccess


ahrtr:
Please take a look at line 428 in pkg/cloud/libvirt/client/client.go,
https://github.com/openshift/cluster-api-provider-libvirt/blob/c01644a170ed2e2cf79ac2fcc55c9131e0c6fac3/pkg/cloud/libvirt/client/client.go#L428

The returned error of waitForSuccess isn't checked, so I think it should be as below,
```
	err = waitForSuccess("error refreshing pool for volume", func() error {
		return client.pool.Refresh(0)
	})
	if err != nil {
		return fmt.Errorf("can't find storage pool '%s'", client.poolName)
	}
```

PR #172, opened 2019-08-26
Catch the error from the function(waitForSuccess) call

ahrtr:
Fix https://github.com/openshift/cluster-api-provider-libvirt/issues/171


zeenix:
Good catch. Commit message could be a bit better and correct though. You're catching the error from the function call, the check was already there. :)

ahrtr:
> Commit message could be a bit better and correct though

Thanks & Commit message updated!

Issue #173, opened 2019-08-30
Ignition configuration injection isn't supported on s390x/mainframe platform


ahrtr:
Currently "fw_cfg" is used by default to inject ignition configuration, but "fw_cfg" isn't supported on s390x/mainframe platform. The proposal is to use guestfish instead to inject ignition configuration for s390x/mainframe.

PR #174, opened 2019-08-30
support injecting ignition configuration for s390x/mainframe platform

ahrtr:
Fix https://github.com/openshift/cluster-api-provider-libvirt/issues/173


jichenjc:
@zeenix can you help to take a look at this PR? Thanks

zeenix:
Sorry, i've been on vacation. My first question is why this is needed? I understand why S390 support is important in general for us but we need to keep in mind that `libvirt` platform (at least currently) is only targeted for developers only. Do we expect S390 to be a developer machine?

ahrtr:
> we need to keep in mind that `libvirt` platform (at least currently) is only targeted for developers. Do we expect S390 to be a developer machine?

@zeenix Yes, we probably may use 'libvirt" in the internal CI & DEV process for s390x platform. But "fw_cfg" isn't supported on s390x/mainframe platform for now, and it's exactly the reason why I submit this pull request.

jichenjc:
per @ahrtr , we decided to run CI on s390x platform with libvirt now, we understand openstack or IBM cloud might be best choices but those are not ready yet ,so we have to start libvirt as current solution, thanks @zeenix

zeenix:
Thanks for explaining. Could you kindly squash the commits so that commits fixing the previous commit, are not separate? As reviewer, I'm only interested in your new changes. :) Once done, i'll review and give the ok-to-test.

ahrtr:
> Once done, i'll review and give the ok-to-test.

Thanks for the comment. Done!

zeenix:
I've also the same question as I asked on the [terraform-provider-libvirt](https://github.com/dmacvicar/terraform-provider-libvirt/pull/629#pullrequestreview-284161107): Is it necessary to rely on guestfs/guestfish here? 

/ok-to-test

ahrtr:
> Is it necessary to rely on guestfs/guestfish here?

It's a good question, and actually I have the same question as well. We need more time to investigate other options, but guestfish works and we will follow this way for now. Probably we can raise another ticket to track this?

jichenjc:
The purpose of the whole flow is to update coreos and specify an ignition to make it consumable
the way x86 is using fw_cfg which is to create a fw_cfg device and append to xml definition of the coreos virtual machine and during startup phase the fw_cfg device will be recognized by driver and ignition handle script will parse the info from fw_cfg device (The ignitino file in it) and do the ignition

because s390x doesn't support fw_cfg and no other alternative to fw_cfg , guestfish is the most popular way to update qcow2 image ,so unless we have other way to inject the ignition file into the qcow2 image I think guestfish is the best way we have...

we might have a way to create a http service and let coreos consume http ignition source but that will increase a requirements on http service so doesn't work in some scenario

zeenix:
> because s390x doesn't support fw_cfg and no other alternative to fw_cfg , guestfish is the most popular way to update qcow2 image 

It's certainly the simplest way of handling this but it has it's downsides, including it being considerably slow (although it's blazing fast compared to rest of the Openshift Installation time) and pulling in OCaml devtools on build systems.

> ,so unless we have other way to inject the ignition file into the qcow2 image I think guestfish is the best way we have...

Ah, `qcow2` image, i missed the fact that you're adding files to that format. Ideally, I'd say you should use another injection method (CDROM image e.g) like GNOME Boxes does for unattended installations but since that requires[ further changes in Installer](https://github.com/openshift/installer/issues/2112) and this code path will only be used for `S390*`, I think it's fine to use guestfish.

pkg/cloud/libvirt/client/domain.go
```diff
@@ -51,44 +51,8 @@ func newDomainDef() libvirtxml.Domain {
...
-					},
-					Target: &libvirtxml.DomainConsoleTarget{
-						Type: "virtio",
-						Port: &serialPort,
-					},
-				},
-			},
-		},
+		CPU:     &libvirtxml.DomainCPU{},
+		Devices: newDevicesDef(),
```
zeenix: Nice!

pkg/cloud/libvirt/client/domain.go
```diff
@@ -105,6 +69,53 @@ func newDomainDef() libvirtxml.Domain {
...
+				},
+				Target: &libvirtxml.DomainConsoleTarget{
+					Type: "virtio",
+					Port: &serialPort,
+				},
+			},
+		},
+	}
+
+	if runtime.GOARCH != "s390x" {
```
zeenix: Why does S390 need a graphic and not others?

pkg/cloud/libvirt/client/ignition_s390x_mock_test.go
```diff
@@ -0,0 +1,264 @@
+package client
```
zeenix: Also a testcase included, nice nice!

pkg/cloud/libvirt/client/ignition.go
```diff
@@ -57,6 +63,199 @@ func setIgnition(domainDef *libvirtxml.Domain, client *libvirtClient, ignition *
 	return nil
 }
 
+func setIgnitionForS390X(domainDef *libvirtxml.Domain, client *libvirtClient, ignition *providerconfigv1.Ignition, kubeClient kubernetes.Interface, machineNamespace, volumeName string) error {
```
zeenix: Quite a bit of code being added for S390 here. Could we please keep it in a separate file? `ignition_s390.go` for example?

pkg/cloud/libvirt/client/domain.go
ahrtr: Graphics/Spice isn't supported on s390x, so I just exclude it from s390x platform.

pkg/cloud/libvirt/client/ignition.go
ahrtr: Good point, but it doesn't work, because "*._s390x.go" or "*_s390.go" will only be complied on s390x platform, accordingly you will get a "undefined: setIgnitionForS390X" error when building the project. 

Two proposals:
1. Use a different name, i.e. "s390x_ignition.go", "ignition_s390a.go", etc..  It works, but I really don't like the names. If you insist, then I would like to use "s390x_igninition.go".
2. Keep it unchanged. 

What do you think?

ahrtr:
> I assume you've tested this to actually work on S390? :)

It's sure thing. I have already verified the change using cmd/libvirt-actuator on s390x/mainframe platform.


> and pulling in OCaml devtools on build systems.
It won't break anything, in other words, you don't need to depend on OCaml devtool on build systems, because we just invoke the guestfish command in runtime.

pkg/cloud/libvirt/client/domain.go
zeenix: ah, i misread the condition. :)

pkg/cloud/libvirt/client/ignition.go
zeenix: Right, good catch. `s390_ignition.go` sounds good to me.

zeenix: or better yet, name it `ignition_s390.go` and name the functions the same `setIgnition` (i-e keep the existing (in master) names), enclose the non-s390 version in a build constraint of `!s390`, and let Go compiler take care of which one to use.

zeenix:
> Ideally, it's the best solution for this kind of "issue". But it won't work for this case, because currently we only support setIgnition() for s390x platform; in other words, we do not support/implement ignKey and cloudInit for s390x platform for now

The functions don't need to do the same things. :) The generic version will just have extra params (if needed). I don't see how this won't work.

zeenix:
> In summary, let's go with the "s390_ignition.go".

The only issue i see with this approach is that it's a bit confusing since you're working around a Go feature with ordering the name. So I'd very much prefer the solution i recommended above but it's no big deal and if you insist we an go your way (at least for now). The major thing is to keep the code separated out IMO that you already did that (thanks!).

pkg/cloud/libvirt/client/client.go
```diff
@@ -425,7 +435,7 @@ func (client *libvirtClient) CreateVolume(input CreateVolumeInput) error {
 		// create the volume
 		// Refresh the pool of the volume so that libvirt knows it is
 		// not longer in use.
-		waitForSuccess("error refreshing pool for volume", func() error {
+		err = waitForSuccess("error refreshing pool for volume", func() error {
```
zeenix: This seems like an unrelated fix (to existing code). Please put it in a separate patch.

pkg/cloud/libvirt/client/domain.go
```diff
@@ -105,6 +69,53 @@ func newDomainDef() libvirtxml.Domain {
...
+				},
+				Target: &libvirtxml.DomainConsoleTarget{
+					Type: "virtio",
+					Port: &serialPort,
+				},
+			},
+		},
+	}
+
+	if runtime.GOARCH != "s390x" || runtime.GOARCH == "s390" {
```
zeenix: I now know why i might have gotten confused about the condition here: it's inverted for `"s390"` by mistake. Please correct it.

zeenix:
Other than those two small things, I think it can go in already.

ahrtr:
> The functions don't need to do the same things. :) The generic version will just have extra params (if needed). I don't see how this won't work.

I got your point now and you are right. But note that some functions (i.e. createAndUpload, createFile, etc.) are common, which will be used by both s390 and non-s390 platform, so if we follow this approach, then we need three files: 
1. ignition.go: for common code
2. s390_ignition.go (do you have a better name?): for s390 or s390x platform;  we need add the following build constrains:
     // +build s390 s390x
3. non_s390_ignition.go (do you have a better name?): for non-390 or non-s390x platform; we need add the following build constrains:
    // +build !s390, !s390x

pkg/cloud/libvirt/client/client.go
ahrtr: Thanks for pointing out:) Actually I did not fix this in this branch manually, instead, I just merged the master branch into this branch before squashing & pushing. Anyway, fixed.

pkg/cloud/libvirt/client/domain.go
ahrtr: Good catch, thanks. Fixed.

ahrtr:
FYI. I just re-ran the unit test and cmd/libvirt-actuator on s390x/mainframe platform, both were OK.

ahrtr:
> Other than those two small things, I think it can go in already.

Done  @zeenix

zeenix:
> Done @zeenix

👍 Please squash and then i'll review.

ahrtr:
> Please squash and then i'll review.

Done.

pkg/cloud/libvirt/client/domain.go
```diff
@@ -105,6 +69,53 @@ func newDomainDef() libvirtxml.Domain {
...
+				},
+				Target: &libvirtxml.DomainConsoleTarget{
+					Type: "virtio",
+					Port: &serialPort,
+				},
+			},
+		},
+	}
+
+	if runtime.GOARCH != "s390x" && runtime.GOARCH != "s390" {
```
zeenix: A short comment here for the s390 check would be great.

ahrtr: > A short comment here for the s390 check would be great.

Done.

zeenix:
Darn, i only noticed now that you also squashed the unrelated change that I asked to be in a separate patch specifically: catch the error from the function(waitForSuccess) call.

pkg/cloud/libvirt/client/client.go
zeenix: it got unfixed when squashing other patches. :(

ahrtr: > it got unfixed when squashing other patches. :(

I did it on purpose. I just want to make sure this pull request do not tangle with any other changes/pull requests.

ahrtr:
> Darn, i only noticed now that you also squashed the unrelated change that I asked to be in a separate patch specifically: catch the error from the function(waitForSuccess) call.

I did it on purpose. I just want to make sure this pull request do not tangle with any other changes/pull requests. Do you want me to merge the master into this branch again?

ahrtr:
> Rebase please and separate out the unrelated change into it's own patch/commit. Thanks.

Thanks & Done.

pkg/cloud/libvirt/client/domain.go
zeenix: I mean the comment saying why Graphics card is not being added for S390. :)

zeenix:
> Thanks & Done.

1. So you went for the approach of putting the change in it's own PR? That's fine but please link to the PR if so.
2. The commit message still has the log from the removed change.

pkg/cloud/libvirt/client/domain.go
ahrtr: Ah:), got it & a new comment added.

ahrtr:
> * So you went for the approach of putting the change in it's own PR? That's fine but please link to the PR if so.
> * The commit message still has the log from the removed change.

Sorry for the confusion. Actually I did not fix any unrelated changes in this PR. Previously I just merged the master into this branch, but the issue was that the change/diff of the other [PR](https://github.com/openshift/cluster-api-provider-libvirt/pull/172) was considered as one of the change/diff of this PR as well after squashing all the commits, and it's exactly the root cause of this confusion. 

Since there aren't any conflicts with the base/master branch, there isn't necessary merge master into this branch, so I did something (I mean the merging from master into this branch) unnecessary. 

In order to remedy the confusion/situation, I rollbacked the merge (excluded the new change from master firstly), and then rebased this branch against master. Now the base of this branch is the HEAD of master, so it includes the new change (https://github.com/openshift/cluster-api-provider-libvirt/pull/172) from master, but it (the new change from master) isn't considered as a change/diff as this PR any more.

Hope it's  clear enough. Please kindly let me know if you have any further comments or concerns.

PR #175, opened 2019-09-10
Add possibility to configure the desired size of the instance volume

cynepco3hahue:
It possible the situation when 16Gb is not enough, to run all needed workloads,
in this situatution the node will have DiskPressure condition and will start
to evict pods from it. We want to prevent such situations and make instance volume
size configurable.

### DIFF
--- a/pkg/apis/libvirtproviderconfig/v1beta1/libvirtmachineproviderconfig_types.go
+++ b/pkg/apis/libvirtproviderconfig/v1beta1/libvirtmachineproviderconfig_types.go
@@ -2,6 +2,7 @@ package v1beta1
 
 import (
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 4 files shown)


zeenix:
@cynepco3hahue Wouldn't this also require changes in Installer?

cynepco3hahue:
@zeenix It will only affect the workers, for the master I hardcodes size value under the terraform `main.tf` file,
```
--- a/data/data/libvirt/main.tf
+++ b/data/data/libvirt/main.tf
@@ -23,6 +23,7 @@ resource "libvirt_volume" "master" {
   count          = "${var.master_count}"
   name           = "${var.cluster_id}-master-${count.index}"
   base_volume_id = "${module.volume.coreos_base_volume_id}"
+  size           = 32212254720
 }
```
I think if installer guys will not be versus it, I can create the installer PR as well, to be honest it is interesting what the default valud for volume size under the AWS.

zeenix:
> I think if installer guys will not be versus it, I can create the installer PR as well, to be honest it is interesting what the default valud for volume size under the AWS.

https://github.com/openshift/installer/issues/2111#issuecomment-517679375

cynepco3hahue:
@zeenix Can we merge it?

cynepco3hahue:
I do not sure that `ci/prow/actuator-pkg-staleness` still relevant after moving to `go mod`, at least I do not see this job anymore under machine-api-operator.

zeenix:
@cynepco3hahue What i was saying was that currently the cpu and ram comes from the Installer through its manifest for worker nodes and you're adding new API to be able to specify volume size too so I'm a bit puzzled as to how this won't require changes in Installer?

cynepco3hahue:
@zeenix Ah, now it more clear:) yes you right it will need changes under the installer as well, but as temp solution, you can generate manifests and `sed` new disk size under the machine set.

cynepco3hahue:
I will create the installer PR on my free time:)

PR #176, opened 2019-10-03
Use go modules for dependencies management

cynepco3hahue:
Use `go mod` for dependency management, it provided by go tools and works better and faster than other dependency managements.

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>


zeenix:
@cynepco3hahue You've no idea how much happy this PR makes me. :)

/lgtm
/approve

Issue #177, opened 2019-10-09
Update Dockerfile and probably related yaml file to support mainframe/s390x platform


ahrtr:
In the following PR, guestfish is used to inject ignition for mainframe/s390x platform. 
https://github.com/openshift/cluster-api-provider-libvirt/pull/174 

So we need to update the Dockerfile so as to get the guestfish tool included in the container image of cluster-api-provider-libvirt. 

Since guestfish operates the OS image file located on the host machine directly, but the cluster-api-provider-libvirt runs inside a POD/container, so the yaml file used to deploy the deployment/POD needs to be updated accordingly so as to get the volume containing the OS image file mounted into the POD/container. If the approach is difficult or unacceptable, then we could need to figure out some other solutions.

jichenjc:
@zeenix can you provide some insight here? Thanks

jichenjc:
I did some PoC ,because we need make the container cluster-api-provider-libvirt able to manage the 
libvirt, so maybe modify the yaml file to add volume then we can see the running libvirt sock and make 

```
[root@gpok159 ~]# docker run -it  -v /var/run:/var/run 15b41b901b99 bash
[root@39bdae7fbac1 /]# ls /var/run/libvirt/libvirt-do
ls: cannot access '/var/run/libvirt/libvirt-do': No such file or directory
[root@39bdae7fbac1 /]# ls /var/run/libvirt/libvirt-sock
/var/run/libvirt/libvirt-sock

[root@4ae28965b635 /]# virsh list
 Id   Name                    State
---------------------------------------
 32   test1-87vbl-bootstrap   running
 34   test1-87vbl-master-0    running

```

so looks like add a volume should be ok to proceed, but not sure we can do so, can you provide some comments? @zeenix

jichenjc:
some PoC on fedora s390 docker images:

[root@gpok159 cluster-api-provider-libvirt]# docker run -it  -v /var/run:/var/run 15b41b901b99 bash
[root@09700f7b1ff6 /]# ls /var/run/libvirt/libvirt-sock
/var/run/libvirt/libvirt-sock

then I install guestfish tools so I can virsh list and run the guestfish command now

next step is to use the container to execute guestfish command to update coreos image..
I guesss /var/run/libvirt/libvirt-sock already mounted into cluster-api-provider-libvirt container for access the libvirt ?

jichenjc:
1) I can use following command to run the docker container
```
[...]

PR #178, opened 2019-10-24
fix s390x create dir issue

jichenjc:
in previous patch #174, we fixed the s390x ignition create
issue by using guestfish, but the dir is not correctly created
this PR added that part

### DIFF
--- a/pkg/cloud/libvirt/client/ignition_s390x_mock_test.go
+++ b/pkg/cloud/libvirt/client/ignition_s390x_mock_test.go
@@ -42,6 +42,9 @@ var supportedCommands = [][]string{
 		"guestfish", "--remote", "--", "mount", "*", "/",
 	},
 	{
+		"guestfish", "--remote", "--", "mkdir-p", "/ignition",
+	},
+	{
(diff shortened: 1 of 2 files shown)


zeenix:
Apart from the commit message being a bit sloppy (no `.` at the end of sentences), sounds good to me. :) Having said that, i'm curious how this was not caught during testing of #174.

/lgtm

jichenjc:
@zeenix to answer your question at #177 , the original image we got from internal does have that folder (the /ignition) so the create success, but after some evolvement the dir is not created anymore in coreos build stage so we have to create dir in this code.. hope that can help the confusion you have, thanks

jichenjc:
updated a new version to fix some problem..

zeenix:
> hope that can help the confusion you have, thanks

Thanks @jichenjc . Indeed it does. :)

zeenix:
> updated a new version to fix some problem..

Thanks. What changed?

jichenjc:
> > updated a new version to fix some problem..
> 
> Thanks. What changed?

sorry, I should be clear in last comment, a unit test failure ...I should run ut locally first next time..

PR #185, opened 2019-12-18
Support config drive for ignition config on PowerPC

Prashanth684:
Following up on #182, the same mechanism is needed for
PowerPC systems for an Openshift baremetal IPI installation
because PowerPC does not support fw_cfg.

Also don't use Spice for the graphics as it is not supported in
PowerPC.

Contributors: @mtarsel

### DIFF
--- a/pkg/cloud/libvirt/client/configdrive_ignition.go
+++ b/pkg/cloud/libvirt/client/configdrive_ignition.go
@@ -17,8 +17,9 @@ import (
 
 var execCommand = exec.Command
 
-func setIgnitionForS390X(domainDef *libvirtxml.Domain, client *libvirtClient, ignition *providerconfigv1.Ignition, kubeClient kubernetes.Interface, machineNamespace, volumeName string) error {
-	glog.Infof("Creating ignition file for s390x")
+// Currently used for s390 and PowerPC arches
+func setIgnitionWithConfigDrive(domainDef *libvirtxml.Domain, client *libvirtClient, ignition *providerconfigv1.Ignition, kubeClient kubernetes.Interface, machineNamespace, volumeName string) error {
+	glog.Infof("Creating ignition file with config drive")
 	ignitionDef := newIgnitionDef()
 
 	if ignition.UserDataSecret == "" {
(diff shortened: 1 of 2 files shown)


pkg/cloud/libvirt/client/domain.go
```diff
@@ -108,7 +108,8 @@ func newDevicesDef(virConn *libvirt.Connect) *libvirtxml.DomainDeviceList {
 	}
 	// Both "s390" and "s390x" are linux kernel architectures for Linux on IBM z Systems, and they are for 31-bit and 64-bit respectively.
 	// Graphics/Spice isn't supported on s390/s390x platform.
-	if arch != "s390x" && arch != "s390" {
+	// Same case for PowerPC systems as well
+	if !strings.HasPrefix(arch, "s390") && !strings.HasPrefix(arch, "ppc64") {
```
zeenix: Fine with me but spice change is a separate logical change so please split the commit (not the PR). Pro tip: If you need to use the word "also" in your commit log, it usually means you've more than one change in your patch. :)

Prashanth684: thanks @zeenix . got it. I was just being lazy ;)

Prashanth684: opened #186

zeenix: Happens to all of us. Please note that I only asked for the commits to be split (not the PR). It's not a problem at all and some might argue it's even better, just for future reference. :)

pkg/cloud/libvirt/client/configdrive_ignition.go
```diff
@@ -2,23 +2,25 @@ package client
...
 	libvirt "github.com/libvirt/libvirt-go"
 	libvirtxml "github.com/libvirt/libvirt-go-xml"
 	providerconfigv1 "github.com/openshift/cluster-api-provider-libvirt/pkg/apis/libvirtproviderconfig/v1beta1"
-	"io"
-	"io/ioutil"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"os"
-	"os/exec"
-	"path/filepath"
```
zeenix: Irrelevant changes, no?

Prashanth684: Ah..my visual studio code editor moved it around..will revert..

PR #187, opened 2020-03-16
Remove bison from OWNERS file

bison:
Removing myself from the OWNERS file.


bison:
I guess this is not going to pass. I'll just close it for now.

PR #188, opened 2020-03-24
Bug 1828961: Use virtio disk instead of config drive for injecting ignition for s390x/ppc64le

Prashanth684:
Similar to https://github.com/dmacvicar/terraform-provider-libvirt/pull/718

The method of mimicking what Openstack does for injecting ignition config works for
images which have the provider as Openstack because ignition recognizes the platform
and knows it has to get the ignition config from the config drive. For QEMU images, ignition
supports getting the config from the firmware config device which is not supported by ppc64
and s390x.

The workaround we have used thus far is to use the Openstack image on the QEMU platform but have this provider create the iso containing the ignition config. There was a discussion in ignition
(coreos/ignition#928) to have a more QEMU based method of injecting ignition config and it was decided to use a virtio-blk device with a serial of ignition which ignition can recognize. This was mainly because with external devices, it is hard to tell if there is an issue with the device or if the kernel has not detected it yet if it has a long discovery phase.

This PR reverts the ISO method used by Openstack and just creates a virtio-blk device through the QEMU command line options.

Reference PR which supports ignition fetching through virtio-blk for QEMU: coreos/ignition#936


zeenix:
The `e2e-libvirt` job won't pass until at least this [PR](https://github.com/openshift/release/pull/7568) is merged.

crawford:
Let’s try this again, now that https://github.com/openshift/release/pull/8592 has merged.

/test e2e-libvirt

crawford:
From the [failure](https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-api-provider-libvirt/188/pull-ci-openshift-cluster-api-provider-libvirt-master-e2e-libvirt/77):

> pods "e2e-dns-utils-q2kss" is forbidden: exec operation is not allowed because the pod's security context exceeds your permissions: pods "e2e-dns-utils-q2kss" is forbidden: securitycontextconstraints.security.openshift.io cache is not synchronized
occurred

/test e2e-libvirt

Prashanth684:
/retitle Bug 1828961: Use virtio disk instead of config drive for injecting ignition for s390x/ppc64le

PR #189, opened 2020-03-27
Bug 1818149: Use virtio disk instead of config drive for injecting ignition for s390x/ppc64le

Prashanth684:
Similar to dmacvicar/terraform-provider-libvirt#718

The method of mimicking what Openstack does for injecting ignition config works for
images which have the provider as Openstack because ignition recognizes the platform
and knows it has to get the ignition config from the config drive. For QEMU images, ignition
supports getting the config from the firmware config device which is not supported by ppc64
and s390x.

The workaround we have used thus far is to use the Openstack image on the QEMU platform but have this provider create the iso containing the ignition config. There was a discussion in ignition
(coreos/ignition#928) to have a more QEMU based method of injecting ignition config and it was decided to use a virtio-blk device with a serial of ignition which ignition can recognize. This was mainly because with external devices, it is hard to tell if there is an issue with the device or if the kernel has not detected it yet if it has a long discovery phase.

This PR replaces the the method of ignition injection from a conig drive disk to a virtio-blk device.

Reference PR which supports ignition fetching through virtio-blk for QEMU: coreos/ignition#936


Prashanth684:
/retitle Bug 1818149: [release-4.3]: Use virtio disk instead of config drive for injecting ignition for s390x/ppc64le

zeenix:
> This PR reverts the ISO method used by Openstack

Openstack? Perhaps a C&P mistake?

Prashanth684:
> > This PR reverts the ISO method used by Openstack
> 
> Openstack? Perhaps a C&P mistake?

The method used previously to support ignition injection for libvirt was to create a config drive disk similar to what Openstack uses and have ignition fetch the config from the disk. This method has now been replaced by a more qemu specific approach.

zeenix:
> The method used previously to support ignition injection for libvirt was to create a config drive disk similar to what Openstack uses and have ignition fetch the config from the disk. This method has now been replaced by a more qemu specific approach.

Thanks, but the commit log/description here isn't clear and seems to claim that this PR revert an Openstack change. Kindly reword it so it's clear.

Prashanth684:
> > The method used previously to support ignition injection for libvirt was to create a config drive disk similar to what Openstack uses and have ignition fetch the config from the disk. This method has now been replaced by a more qemu specific approach.
> 
> Thanks, but the commit log/description here isn't clear and seems to claim that this PR revert an Openstack change. Kindly reword it so it's clear.

Understood. I changed that line to: 

`This PR replaces the the method of ignition injection from a conig drive disk to a virtio-blk device which is specified through the QEMU command line options.`

pkg/cloud/libvirt/client/domain_test.go
```diff
@@ -28,7 +29,7 @@ TestCases:
 	for i, tc := range testCases {
 		domainDef := libvirtxml.Domain{}
 
-		err := setCoreOSIgnition(&domainDef, tc.ignKey)
+		err := setCoreOSIgnition(&domainDef, tc.ignKey, runtime.GOARCH)
```
zeenix: why does arch needs to be passed from here? Can't `setCoreOSIgnition` access the `runtime.GOARCH` itself directly?

Prashanth684: The `setCoreOSIgnition` function needs the arch variable in its args as it gets it from `getHostArchitecture`. So for the test, this needs to be passed in.

crawford:
/override ci/prow/actuator-pkg-staleness

This test isn't applicable to this repo (we don't use Go Modules).

crawford:
Apparently overrides aren't enough to stave off @openshift-bot.

/hold

zeenix:
> Apparently overrides aren't enough to stave off @openshift-bot.

But `actuator-pkg-staleness` isn't a required test and shouldn't block the merge so not sure what's happening here.

crawford:
> But `actuator-pkg-staleness` isn't a required test and shouldn't block the merge so not sure what's happening here.

The merge was blocked because this patch is against 4.3 and we only merge into release branches during specific times. Even though the test wasn't required, the bot continued to retest, which was just needlessly using up resources. Now that the release window has opened, I added the cherry-pick-approved label, which will allow this to merge.

PR #194, opened 2020-05-19
ignition: use setCoreOSIgnition function to inject ignition config

Prashanth684:
A minor refactoring of code which is repeated in domain.go. The setCoreOSIgnition
function is handling the injection of ignition. There is no need to repeat this code as this function
can just be used.


cfergeau:
Looks good to me.

Prashanth684:
/retitle Bug 1840695: ignition: use setCoreOSIgnition function to inject ignition config

PR #195, opened 2020-05-19
Stop building unused binary

cfergeau:


cfergeau:
The `fixup` commit is intentional, I've been told that on merge, github would automatically squash such commits, but I never tried it. I can squash manually if you prefer :)

PR #199, opened 2020-07-08
BUG 1855288: fix memory leak that leads to unresponsive libvirt URI

dbenoit17:
The libvirt C library will maintain open connections against the URI based on an internal reference counting scheme.  This pull request adds calls to Free() for a storage pool and a network object which otherwise cause unused connections to linger and stack, causing an unresponsive URI at scale.

### DIFF
--- a/pkg/cloud/libvirt/client/client.go
+++ b/pkg/cloud/libvirt/client/client.go
@@ -155,9 +155,14 @@ func NewClient(URI string, poolName string) (Client, error) {
 
 // Close closes the client's libvirt connection.
 func (client *libvirtClient) Close() error {
-	glog.Infof("Closing libvirt connection: %p", client.connection)
+	glog.Infof("Freeing the client pool")
+	err := client.pool.Free()
+	if err != nil {
+		glog.Infof("Error freeing the client pool: %v", err)
+	}
 
-	_, err := client.connection.Close()
+	glog.Infof("Closing libvirt connection: %p", client.connection)
+	_, err = client.connection.Close()
 	if err != nil {
 		glog.Infof("Error closing libvirt connection: %v", err)
 	}
@@ -558,6 +563,7 @@ func (client *libvirtClient) GetDHCPLeasesByNetwork(networkName string) ([]libvi
 		glog.Errorf("Failed to fetch network %s from the libvirt", networkName)
 		return nil, err
 	}
+	defer network.Free()
 
 	return network.GetDHCPLeases()
 }
--- a/pkg/cloud/libvirt/client/client.go
+++ b/pkg/cloud/libvirt/client/client.go
@@ -196,6 +196,7 @@ func (client *libvirtClient) CreateDomain(ctx context.Context, input CreateDomai
 	if err != nil {
 		return fmt.Errorf("can't retrieve volume %s for pool %s: %v", input.VolumeName, client.poolName, err)
 	}
+	defer diskVolume.Free()
 	if err := setDisks(&domainDef, diskVolume); err != nil {
 		return fmt.Errorf("Failed to setDisks: %s", err)
 	}
@@ -216,6 +217,7 @@ func (client *libvirtClient) CreateDomain(ctx context.Context, input CreateDomai
 		if err != nil {
 			return fmt.Errorf("error getting ignition volume: %v", err)
 		}
+		defer ignVolume.Free()
 		ignVolumePath, err := ignVolume.GetPath()
 		if err != nil {
 			return fmt.Errorf("error getting ignition volume path: %v", err)
@@ -387,6 +389,7 @@ func (client *libvirtClient) CreateVolume(input CreateVolumeInput) error {
 
 	volume, err := client.getVolume(input.VolumeName)
 	if err == nil {
+		volume.Free()
 		return fmt.Errorf("storage volume '%s' already exists", input.VolumeName)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/cloud/libvirt/client/client.go
```diff
@@ -559,7 +565,14 @@ func (client *libvirtClient) GetDHCPLeasesByNetwork(networkName string) ([]libvi
 		return nil, err
 	}
 
-	return network.GetDHCPLeases()
+	dhcpLeases, err := network.GetDHCPLeases()
+	if err != nil {
+		glog.Errorf("Network %s failed to return dhcp leases", networkName)
+		return nil, err
+	}
+	network.Free()
```
Prashanth684, line 573: should we just do a `defer network.Free()` when we get the network object above after the nil sanity check ?

dbenoit17: oh yeah, good call.  the original patch doesnt cover the second error case

Prashanth684:
/cc @praveenkumar @cfergeau 
This is needed for our multi-arch CI which runs on libvirt. We were noticing that after running a few jobs the libvirt daemon would hang because of these open connections.

pkg/cloud/libvirt/client/client.go
```diff
@@ -162,6 +162,12 @@ func (client *libvirtClient) Close() error {
 		glog.Infof("Error closing libvirt connection: %v", err)
 	}
 
+	glog.Infof("Freeing the client pool")
```
cfergeau, line 165: Since the storage pool keeps a reference on the libvirt connection, I would free it before closing the connection in case it needs to use the connection to do some cleanups.

Prashanth684:
> Looking over this file, it seems there are more places where we leak pools
> 
> ```
> diff --git a/pkg/cloud/libvirt/client/client.go b/pkg/cloud/libvirt/client/client.go
> index 7ff8af70..1aa2a546 100644
> --- a/pkg/cloud/libvirt/client/client.go
> +++ b/pkg/cloud/libvirt/client/client.go
> @@ -194,6 +194,7 @@ func (client *libvirtClient) CreateDomain(ctx context.Context, input CreateDomai
>         if err != nil {
>                 return fmt.Errorf("can't retrieve volume %s for pool %s: %v", input.VolumeName, client.poolName, err)
>         }
> +       defer diskVolume.Free()
>         if err := setDisks(&domainDef, diskVolume); err != nil {
>                 return fmt.Errorf("Failed to setDisks: %s", err)
>         }
> @@ -214,6 +215,7 @@ func (client *libvirtClient) CreateDomain(ctx context.Context, input CreateDomai
>                 if err != nil {
>                         return fmt.Errorf("error getting ignition volume: %v", err)
>                 }
> +               defer ignVolume.Free()
>                 ignVolumePath, err := ignVolume.GetPath()
>                 if err != nil {
>                         return fmt.Errorf("error getting ignition volume path: %v", err)
> @@ -387,6 +389,8 @@ func (client *libvirtClient) CreateVolume(input CreateVolumeInput) error {
>         if err == nil {
>                 return fmt.Errorf("storage volume '%s' already exists", input.VolumeName)
>         }
> +       /* probably not quite right in the input.BaseVolumeName == "" case */
> +       defer volume.Free()
>  
>         volumeDef := newDefVolume(input.VolumeName)
>         volumeDef.Target.Format.Type = input.VolumeFormat
> @@ -417,6 +421,7 @@ func (client *libvirtClient) CreateVolume(input CreateVolumeInput) error {
>                 if err != nil {
>                         return fmt.Errorf("Can't retrieve volume %s", input.BaseVolumeName)
>                 }
> +               defer baseVolume.Free()
>                 var baseVolumeInfo *libvirt.StorageVolInfo
>                 baseVolumeInfo, err = baseVolume.GetInfo()
>                 if err != nil {
> ```
> 
> (the `CreateVolume` one probably can't be used as is).
> 
> This PR is mentioning memory leaks, but I was under the impression that the main issue was that too many sockets to libvirtd were opened? Or was the number of sockets just a symptom, and the real problem the amount of memory consumed on the system?

There would be socket objects lingering because the memory was not freed, so the sockets would be kept open. Freeing this memory also closed the socket. So, in that sense it is a leak. But in terms of "memory" as such I think it was fine. So, these memory leaks are good to cleanup, but might not impact as much.

cfergeau:
> There would be socket objects lingering because the memory was not freed, so the sockets would be kept open. Freeing this memory also closed the socket. So, in that sense it is a leak. But in terms of "memory" as such I think it was fine. So, these memory leaks are good to cleanup, but might not impact as much.

Yes, freeing the object will release the references they keep to the libvirt connection, which will allow this connection to be freed, which will close the sockets. If the issue really was that too many sockets were opened, it would be useful to mention this in the commit log.

pkg/cloud/libvirt/client/client.go
cfergeau: Actually, it's not that important because `virConnectClose` is mostly `virConnectUnref`, so the connection won't be really closed until the storage pool is freed too (which is what this bug is about after all :)
One nice thing about freeing the pool first is that we can check `client.connection.Close()` return value to warn about leaks in the future https://libvirt.org/html/libvirt-libvirt-host.html#virConnectClose

> a positive number if at least 1 reference remains on success. The returned value should not be assumed to be the total reference count. A return of 0 implies no references remain and the connection is closed and memory has been freed.

dbenoit17: This is really good to know!

cfergeau: I added this in https://github.com/cfergeau/cluster-api-provider-libvirt/commits/leaks

cfergeau:
> > There would be socket objects lingering because the memory was not freed, so the sockets would be kept open. Freeing this memory also closed the socket. So, in that sense it is a leak. But in terms of "memory" as such I think it was fine. So, these memory leaks are good to cleanup, but might not impact as much.
> 
> Yes, freeing the object will release the references they keep to the libvirt connection, which will allow this connection to be freed, which will close the sockets. If the issue really was that too many sockets were opened, it would be useful to mention this in the commit log.

Can we have a real commit log for the 'leak fixes' commit which goes along the lines of this thread?

PR #213, opened 2020-11-30
Support firmware and NVRAM options for UEFI architectures

Prashanth684:
Add support to specify firmware and NVRAM file for UEFI architectures like aarch64. This is very similar to what is done
in the terraform-provider repo [1]. This will allow for these options to be specified in the terraform files in the openshift
installer and enable libvirt machine creation.

[1] https://github.com/dmacvicar/terraform-provider-libvirt/commit/44a504bcc42e16a6f7d8589fe23f602581838d8a


Prashanth684:
@praveenkumar @cfergeau some context here - we are working on enabling ARM support in the near future, so this would be needed. I have tested this with relevant changes to the installer code on an aarch64 system.

cfergeau (commented):
My understanding is that libvirt will be able to pick up a suitable default for firmware/nvram most of the time, even on aarch64? But I'll admit I'm not familiar with aarch64 virt :) Can you describe a bit more the issue this is solving? What would be using this libvirt aarch64 support? Baremetal deployments? Or just testing envs?

Prashanth684:
> My understanding is that libvirt will be able to pick up a suitable default for firmware/nvram most of the time, even on aarch64? But I'll admit I'm not familiar with aarch64 virt :) Can you describe a bit more the issue this is solving? What would be using this libvirt aarch64 support? Baremetal deployments? Or just testing envs?

So, the `virt-install` command has an option `--boot firmware=efi` which will let it know that it is EFI and virt-install will generate the xml with the appropriate path to the firmware file, but in our case we have to populate the xml with the EFI file path. The NVRAM file on the other hand is optional. In `/etc/libvirt/qemu.conf` there is an NVRAM section which has the defaults which can be enabled so that it will pick the correct file. I included it for completeness (mimicking the terraform provider) and in case we need it in the future. So, this change will allow us to make changes to the installer terraform files to specify the firmware file which will be used to create the VM. 

As for the problem we are solving - there will be three deployments of ARM:
- on AWS
- baremetal
- libvirt
right now we are using a UPI solution for libvirt which is cumbersome to setup and test deploys quickly. Which is why it would be good to have IPI support - mainly for quick testing and a lightweight CI solution.

pkg/cloud/libvirt/client/domain.go
```diff
@@ -109,7 +109,10 @@ func newDevicesDef(virConn *libvirt.Connect) *libvirtxml.DomainDeviceList {
 	// Both "s390" and "s390x" are linux kernel architectures for Linux on IBM z Systems, and they are for 31-bit and 64-bit respectively.
 	// Graphics/Spice isn't supported on s390/s390x platform.
 	// Same case for PowerPC systems as well
-	if !strings.HasPrefix(arch, "s390") && !strings.HasPrefix(arch, "ppc64") {
+	switch arch {
+	case "s390", "s390x", "ppc64", "ppc64le", "aarch64":
```
yselkowitz, line 113: On which architectures *is* SPICE supported?  I'm wondering if this shouldn't be changed so that new architectures don't need to continually be added here.

Prashanth684: i thought it's better to blacklist so we don't miss any arches. https://www.spice-space.org/faq.html claims support for arm, but libvirt still threw an error saying SPICE graphics is not supported for qemu, so not sure about that.

But yes - this is not ideal - maybe it would be better to just enable it for x86_64. @cfergeau thoughts?

cfergeau: You have upstream support, and Red Hat support, which can be different things. On RHEL, I see that it's built for x86_64 and aarch64. I think there has been work to make spice support optional, so it may depend on which packages are installed.
Since we also add a console device just before the spice graphics element, it's probably fine to only add it on x86_64 for now. If the console is not enough for debugging purposes, in the future, we can add a vnc graphics element on platforms where spice is disabled.

cfergeau:
> So, the virt-install command has an option --boot firmware=efi which will let it know that it is EFI and virt-install will generate the xml with the appropriate path to the firmware file, but in our case we have to populate the xml with the EFI file path.

I'm wondering if `<os firmware='efi'>` https://libvirt.org/formatdomain.html#bios-bootloader could be used to trigger the same autoselection mechanism as `virt-install`? I'm not opposed to adding paths to the firmware/nvram if that's required, but ideally the simpler solution would be enough?

pkg/cloud/libvirt/client/domain.go
Prashanth684: yup..so looks like qemu-kvm in RHEL is not compiled with spice for non-x86..just looked at the spec file in the rpm:
```
%ifarch x86_64
    %global kvm_target    x86_64
%else
    %global have_spice   0
    %global have_opengl  0
    %global have_gluster 0
```

i can put a request to have it enabled for aarch64, but in the meanwhile i think I will change the conditional here to only support it only for x86.

Prashanth684:
> > So, the virt-install command has an option --boot firmware=efi which will let it know that it is EFI and virt-install will generate the xml with the appropriate path to the firmware file, but in our case we have to populate the xml with the EFI file path.
> 
> I'm wondering if `<os firmware='efi'>` https://libvirt.org/formatdomain.html#bios-bootloader could be used to trigger the same autoselection mechanism as `virt-install`? I'm not opposed to adding paths to the firmware/nvram if that's required, but ideally the simpler solution would be enough?

Yes! thanks for this suggestion. Turns out the firmware attribute is supported in newer libvirt libraries. I tested out and it works well. I have opened #214. If you think those changes are good i will close this PR.

pkg/cloud/libvirt/client/domain.go
Prashanth684: i will open a separate PR for this

Prashanth684:
closing this as i think #214 and #215 address all the issues

PR #214, opened 2020-12-03
Set firmware attribute in libvirt domain for aarch64

Prashanth684:
For aarch64, we need to specify the firmware to be efi so the loader and the nvram is correctly
populated with the appropriate files for machine creation.

This actually started in #213 similar to what was done in the terraform provider, but the newer
libvirt-go-xml libraries support setting of this attribute making things easier


pkg/cloud/libvirt/client/domain.go
```diff
@@ -460,35 +460,43 @@ type Config struct {
...
 
-func domainDefInit(domainDef *libvirtxml.Domain, name string, memory, vcpu int) error {
-	if name != "" {
-		domainDef.Name = name
+func setFirmware(input *CreateDomainInput, domainDef *libvirtxml.Domain, arch string) {
+	// for aarch64 speciffying this will automatically select the firmware and NVRAM file
+	// reference: https://libvirt.org/formatdomain.html#bios-bootloader
+	if arch == "aarch64" {
+		domainDef.OS.Firmware = "efi"
+	}
```
Prashanth684, line 468: I am ok with hardcoding this for aarch64, but do we want to expose this through the machine provider api so it can be set through the installer and this code can be arch agnostic?

Prashanth684: ok..on second thought .. probably not...this is a basic property of aarch64...better to leave it as is and keep things simple

cfergeau: It must not be so basic, otherwise it would be the libvirt default ;) I agree with leaving this as is for now, as I don't think cluster-api-provider-libvirt should support arbitrary VM configurations. This is one case where it should be able to pick up the correct value needed by openshift. For x86_64, the efi switch would happen when we start using the q35 machine type.

go.sum
```diff
@@ -295,10 +295,10 @@ github.com/kr/pty v1.1.5/go.mod h1:9r2w37qlBe7rQ6e1fg1S/9xpWHSnaqNdHD3WcMdbPDA=
 github.com/kr/text v0.1.0 h1:45sCR5RtlFHMR4UwH9sdQ5TC8v0qDQCHnXt+kaKSTVE=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
 github.com/kubernetes-sigs/kube-storage-version-migrator v0.0.0-20191127225502-51849bc15f17/go.mod h1:enH0BVV+4+DAgWdwSlMefG8bBzTfVMTr1lApzdLZ/cc=
-github.com/libvirt/libvirt-go v4.6.0+incompatible h1:7OenwWTPRPSYIkJMMSX9loYLBBOaGdOjxZHTkXSe67E=
-github.com/libvirt/libvirt-go v4.6.0+incompatible/go.mod h1:34zsnB4iGeOv7Byj6qotuW8Ya4v4Tr43ttjz/F0wjLE=
-github.com/libvirt/libvirt-go-xml v4.6.0+incompatible h1:dL2JDzkEtsiZkPh42KHvLt208wgNfO40IsBjHtaXohs=
-github.com/libvirt/libvirt-go-xml v4.6.0+incompatible/go.mod h1:oBlgD3xOA01ihiK5stbhFzvieyW+jVS6kbbsMVF623A=
+github.com/libvirt/libvirt-go v5.10.0+incompatible h1:01fwkdUHH2hk4YyFNCr48OvSGqXYLzp9cofUpeyeLNc=
+github.com/libvirt/libvirt-go v5.10.0+incompatible/go.mod h1:34zsnB4iGeOv7Byj6qotuW8Ya4v4Tr43ttjz/F0wjLE=
```
cfergeau, line 299: I don't think it's required to bump `libvirt-go`, only `libvirt-go-xml` should be needed? If you bump the `libvirt-go` dependency, this can cause unexpected failures to run the installer (if you build on a system where libvirt 5.10 is installed and you then try to run it on a system with libvirt 4.6 installed). I'd prefer to keep the libvirt-go dependency unchanged if possible.

Prashanth684: i changed libvirt-go and libvirt-go-xml to match that of the openshift installer. I thought it would be good to keep both in sync? in any case the installer uses the same versions so should be fine ?

cfergeau: I think the installer upgraded to a newer version indirectly (it updated another module which needed a newer version). This caused issues fwiw https://github.com/openshift/installer/pull/4339
Since we already bit the bullet with the installer, we can try changing it here as well :)

cfergeau (approved):

PR #218, opened 2021-03-11
Bug 1936488 : Update MAO and set metrics on :8081 address

Prashanth684:
This is to comply with https://github.com/openshift/machine-api-operator/pull/609. Without this change alerts are firing as nobody is listening on port 8081 and the metrics can't be exported.

Co-authored-by: andymcc <amccrae@redhat.com>


Prashanth684:
> /lgtm

thanks! mind dropping an approval too? :)

Prashanth684:
@yselkowitz @andymcc  and i discussed this.we need manual backports as we need to match the version of MAO and k8s libs used in 4.7 with other providers rather than just applying this blindly. I am working on this.

PR #219, opened 2021-03-12
Bug 1938316: [release-4.7] Update MAO and set metrics on :8081 address

Prashanth684:
same as #218 for 4.7.


Prashanth684:
i believe the below fix would be needed for e2e-libvirt although the current run's failure is related to some account issue.

https://github.com/openshift/release/pull/16817

Prashanth684:
@praveenkumar i believe we need this:  https://github.com/openshift/release/pull/16817 to merge before the e2e-libvirt can pass.

PR #230, opened 2021-11-02
Bug 2018517: Create event only if the machine was modified

Fedosin:
Now Libvirt actuator always creates an event after Update, even though the change might not have happened.

To prevent creating fake events we check if the update happens for real and create an event only if it's true.


Fedosin:
Currently blocked by https://github.com/openshift/release/pull/23266

PR #235, opened 2022-03-10
Bug 2018517: fix the check that machine has been modified

Fedosin:
Now we convert new machine status into raw status (basically a byte array) and then compare it with the raw status of the existing machine.

Unfortunately, encoder doesn't preserve the order of status fields, which leads to the fact that even if statuses are the same, their raw representations are different.

To fix this we convert raw machine status to a golang object, and then compare it with the provided one. Additionally we have to compare addresses if they were provided.

### DIFF
--- a/pkg/cloud/libvirt/actuators/machine/actuator.go
+++ b/pkg/cloud/libvirt/actuators/machine/actuator.go
@@ -401,14 +401,25 @@ func (a *Actuator) applyMachineStatus(
 	status *providerconfigv1.LibvirtMachineProviderStatus,
 	addrs []corev1.NodeAddress,
 ) (bool, error) {
-	// Encode the new status as a raw extension.
-	rawStatus, err := EncodeProviderStatus(a.codec, status)
+	machineCopy := machine.DeepCopy()
+
+	// Check if machine status has been modified.
+	machineStatus, err := ProviderStatusFromMachine(a.codec, machine)
 	if err != nil {
 		return false, err
 	}
 
-	machineCopy := machine.DeepCopy()
-	machineCopy.Status.ProviderStatus = rawStatus
+	// We need to update provider status only if it is nil (when we create new machines), or if
+	// the status has been modified.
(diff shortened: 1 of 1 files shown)


JoelSpeed (commented):
/approve

pkg/cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -408,7 +408,15 @@ func (a *Actuator) applyMachineStatus(
...
 	machineCopy := machine.DeepCopy()
-	machineCopy.Status.ProviderStatus = rawStatus
+
+	// Check if machine status has been modified.
+	machineStatus, err := ProviderStatusFromMachine(a.codec, machine)
+	if err != nil {
+		return false, err
+	}
+	if status != nil || !equality.Semantic.DeepEqual(machineStatus, status) {
+		machineCopy.Status.ProviderStatus = rawStatus
```
cfergeau, line 418: Can you move `rawStatus, err := EncodeProviderStatus(a.codec, status)` to this block as this is the only place where this is used?

pkg/cloud/libvirt/actuators/machine/actuator.go
```diff
@@ -408,7 +408,15 @@ func (a *Actuator) applyMachineStatus(
...
 
 	machineCopy := machine.DeepCopy()
-	machineCopy.Status.ProviderStatus = rawStatus
+
+	// Check if machine status has been modified.
+	machineStatus, err := ProviderStatusFromMachine(a.codec, machine)
+	if err != nil {
+		return false, err
+	}
+	if status != nil || !equality.Semantic.DeepEqual(machineStatus, status) {
```
cfergeau, line 417: I'm unsure about the `if status != nil || ` part of the check.
It currently is: "If status is not `nil`, then set `machineCopy.Status.ProviderStatus` to `rawStatus`" or "if status is `nil`, then compare status against `machineStatus`"

PR #236, opened 2022-04-21
Updating ose-libvirt-machine-controllers images to be consistent with ART

praveenkumar:
This should supersede #233


cfergeau (approved):
The pkg/apis/libvirtproviderconfig/v1beta1/zz_generated.deepcopy.go changes will be overwritten if something/someone runs `make generate`, but this should do  for now.

PR #249, opened 2023-01-31
Update k8s/vendoring/...

cfergeau:
This updates cluster-api-provider-libvirt to use more recent dependencies, and should help with build problems some people have been hitting such as https://github.com/openshift/installer/pull/6812

`make test` does not pass after these changes, I did not have time to investigate why. This PR should help us get more extensive testing through CI to see exactly how much is non-working.


praveenkumar:
#251 is taking all those patches and add another commit to disable actuator unit tests.

PR #263, opened 2023-09-22
Add workaround for SLOF regression

cfergeau:
A SLOF regression in rhel 8.6 prevents OpenShift libvirt worker nodes from booting.
It's easy to workaround if the VM has a serial (non-virtio) device, so this PR tries to add one.


cfergeau:
I tested this on a beaker ppc64le machine, and the worker nodes could get an IP, and I could ssh into them.

gbraad (approved):

mkumatag:
/retitle "OCPBUGS-17476: Add workaround for SLOF regression"

PR #269, opened 2023-09-29
OCPBUGS-19926:[release-4.14] Don't force use of virtio console

cfergeau:
When adding a console to libvirt XML definition, it's a virtio-console
which is added. This triggered an unrelated regression on ppc64le
https://issues.redhat.com/browse/OCPBUGS-17476

This commit lets libvirt uses its preferred device for the console
depending on the platform. On ppc64le, this should correspond to spapr
devices.

This should workaround https://issues.redhat.com/browse/OCPBUGS-17476


cfergeau:
This supersedes https://github.com/openshift/cluster-api-provider-libvirt/pull/265

PR #294, opened 2025-11-13
mockgen deprecated: use uber-go/mock instead

sebrandon1:
https://github.com/golang/mock is marked as archived as of `June 27, 2023`.  They recommend using go.uber.org/mock.

This PR attempts to change the dependency to one that is maintained.

Tracking issue: https://github.com/redhat-best-practices-for-k8s/telco-bot/issues/45


cfergeau:
I believe cluster-api-provider-libvirt is due for retirement, I’m not sure there are any users of it, and my understanding is that we don’t support openshift clusters running on libvirt? 
If that’s correct, I don’t think it’s worth making large changes such as this one.

sebrandon1:
No worries!  Closing!

PR #301, opened 2025-12-02
[release-4.17] OCPBUGS-77195: Fix incomplete vendor/

adobes1:
Hermetic builds require that running `go mod vendor` produces no diff. Without this PR, ose-libvirt-machine-controllers can't be built hermetically.

The changes in this PR are the result of running go mod vendor and ensuring that the vendored file is commited and not ignored by `.gitignore`


sdodson:
/remove-jira/invalid-bug

PR #302, opened 2025-12-02
[release-4.16] OCPBUGS-77196: Fix incomplete vendor/

adobes1:
Hermetic builds require that running `go mod vendor` produces no diff. Without this PR, ose-libvirt-machine-controllers can't be built hermetically.

The changes in this PR are the result of running `go mod vendor` and ensuring that the vendored file is commited and not ignored by `.gitignore`


cfergeau:
This has some `go.sum` changes, it’s the only one among all the PRs which were opened, dunno if that’s expected.

PR #308, opened 2026-02-24
[release-4.21] OCPBUGS-77191: Fix incomplete vendor/

adobes1:
Hermetic builds in Konflux require that running `go mod vendor` produces no diff. Without this PR, ART can't build `ose-libvirt-machine-controllers ` hermetically.

The changes in this PR are the result of running `go mod vendor` and ensuring that the vendored file is commited and not ignored by `.gitignore`


sdodson:
/close
we're removing this image in https://issues.redhat.com/browse/ART-14723

